PR #3, opened 2015-02-27
Review for DM-1576

TallJimbo:


include/lsst/meas/astrom/detail/utils.h
```diff
@@ -5,21 +5,23 @@
...
-        std::string magcol;
-        std::string magerrcol;
+    struct MagColInfo {
+        std::string filterName; ///< name of filter
+        std::string magCol;     ///< name of magnitude column
+        std::string magErrCol;  ///< name of magnitude sigma column
 
         bool hasErr() const {
-            return (magerrcol.size() > 0);
+            return (magErrCol.size() > 0);
```
TallJimbo: Since you're changing this line anyway, maybe use `!magErrCol.empty()`?

include/lsst/meas/astrom/detail/utils.h
```diff
@@ -60,18 +62,53 @@ struct IndexManager {
...
-               double ra, double dec, double radius,
-               const char* idcol,
-               std::vector<mag_column_t> const& magcols,
-               const char* stargalcol,
-               const char* varcol,
-               bool unique_ids);
+getCatalogImpl(
+    std::vector<index_t*> inds,
+    lsst::afw::coord::Coord const &ctrCoord,
+    lsst::afw::geom::Angle const &radius,
```
TallJimbo: These `lsst::`s are unnecessary, as you're already within the LSST namespace (and unlike other namespace components we have, I don't think there's any need to disambiguate `afw`).

python/lsst/meas/astrom/astrometry.py
```diff
@@ -0,0 +1,257 @@
...
+        dtype = bool,
+        doc= "Assume that the input image's WCS is correct, without comparing it to any external reality " +
+            "  (but still match reference objects to sources)",
+        default = False,
+    )
+
+class AstrometryTask(pipeBase.Task):
+    """!Match input sourceCat with a reference catalog and solve for the Wcs
+
+    \anchor AstrometryTask_
```
TallJimbo: You're using `\` for Doxygen tags in this docstring, but the file uses `@` everywhere else.  I have a slight preference for `@` (because it uniquely identifies markup as Doxygen), but I think it's more important to get consistency within a file one way or the other.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -0,0 +1,257 @@
...
+        debug = lsstDebug.Info(__name__)
+
+        loadRes = self.refObjLoader.loadObjectsInBBox(
+            bbox = bbox,
+            wcs = initWcs,
+            filterName = filterName,
+            calib = calib,
+        )
+        if debug.display:
+            frame = debug.frame if isinstance(debug.frame, int) else 1
```
TallJimbo: I don't particularly like the `isinstance` check here (especially since Python 2 also has `long`, which isn't a subclass of `int`); what could `debug.frame` be other than an integer or `None`?

RobertLuptonTheGood: Or bool

python/lsst/meas/astrom/astrometry.py
```diff
@@ -0,0 +1,257 @@
...
+        - refCat  reference object catalog of objects that overlap the exposure (with some margin)
+            (an lsst::afw::table::SimpleCatalog)
+        - matchList  list of reference object/source matchList (an lsst.afw.table.ReferenceMatchVector)
+        - initWcs  initial WCS from exposure (possibly tweaked) (an lsst.afw.image.Wcs)
+        - wcs  fit WCS (an lsst.afw.image.Wcs)
+        """
+        import lsstDebug
+        debug = lsstDebug.Info(__name__)
+
+        loadRes = self.refObjLoader.loadObjectsInBBox(
```
TallJimbo: I'd prefer just spelling out `Result` here instead of `Res`.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -0,0 +1,257 @@
...
+    wcsFitter = pexConfig.ConfigurableField(
+        target = FitTanSipWcsTask,
+        doc = "WCS fitter",
+    )
+    forceKnownWcs = pexConfig.Field(
+        dtype = bool,
+        doc= "Assume that the input image's WCS is correct, without comparing it to any external reality " +
+            "  (but still match reference objects to sources)",
+        default = False,
+    )
```
TallJimbo: Since it now means "don't run the fitter", how about changing the name of "forceKnownWcs"?  Maybe "doMatchOnly" or "useInitialWcs"?

python/lsst/meas/astrom/astrometry.py
```diff
@@ -0,0 +1,257 @@
...
+        if debug.display:
+            showAstrometry(refCat=loadRes.refCat, sourceCat=sourceCat, matchList=matchRes.matchList,
+                exposure=exposure, bbox=bbox, frame=frame + 2, title="TAN-SIP WCS")
+
+        return pipeBase.Struct(
+            refCat = loadRes.refCat,
+            matchList = matchRes.matchList,
+            initWcs = initWcs,
+            wcs = fitWcs,
+        )
```
TallJimbo: Newlines in this function just seem a little arbitrary.

python/lsst/meas/astrom/astrom.py
```diff
@@ -244,7 +249,7 @@ def _getImageParams(self, wcs, exposure, filterName=None, imageSize=None,
             y0 = 0
         return filterName, imageSize, x0, y0
 
-    def useKnownWcs(self, sources, wcs=None, exposure=None, filterName=None, imageSize=None, x0=None, y0=None):
+    def useKnownWcs(self, sourceCat, wcs=None, exposure=None, filterName=None, imageSize=None, x0=None, y0=None):
```
TallJimbo: Longer than 110 columns.

include/lsst/meas/astrom/matchOptimisticB.h
```diff
@@ -0,0 +1,137 @@
...
+#include "lsst/afw/table/Source.h"
+#include "lsst/afw/table/Match.h"
+
+namespace lsst {
+namespace meas {
+namespace astrom {
+
+    /**
+    A wrapper around a SimpleRecord that allows us to associate an pixel position
+    with a position reference star.
```
TallJimbo: "an pixel position"

include/lsst/meas/astrom/matchOptimisticB.h
```diff
@@ -0,0 +1,137 @@
...
+
+    RecordProxy can also wrap a SourceRecord (in which case the pixel position in the proxy
+    should match that in the record).
+    */
+    struct RecordProxy {
+        PTR(lsst::afw::table::SimpleRecord) record;
+        lsst::afw::geom::Point2D position;
```
TallJimbo: I thought you were going to get rid of `RecordProxy` in favor of having a scratch-space centroid in the schema returned by the loader interface.  Did that turn out not to work, or have you just not gotten there yet?

include/lsst/meas/astrom/matchOptimisticB.h
```diff
@@ -0,0 +1,137 @@
...
+            minMatchedPairs(50),
+            matchingAllowancePix(10.0),
+            maxOffsetPix(300),
+            maxRotationRad(0.02),
+            angleDiffFrom90(0.05),
+            numPointsForShape(6),
+            maxDeterminant(0.02)
+            {
+                validate();
+            }
```
TallJimbo: Indentation for braces seems weird here.

include/lsst/meas/astrom/matchOptimisticB.h
```diff
@@ -0,0 +1,137 @@
...
+        lsst::afw::table::SimpleCatalog const &posRefCat,
+        lsst::afw::table::SourceCatalog const &sourceCat,
+        MatchOptimisticBControl const &control,
+        int posRefBegInd=0,
+        bool verbose = false
+    );
+
+    }
+  }
+}
```
TallJimbo: Close braces should not be indented differently (I'd have just put them all on the same line myself).

python/lsst/meas/astrom/anetAstrometry.py
```diff
@@ -0,0 +1,464 @@
...
+     - \ref pipe_tasks_astrometry_Config
+     - \ref pipe_tasks_astrometry_Debug
+     - \ref pipe_tasks_astrometry_Example
+
+    \section pipe_tasks_astrometry_Purpose	Description
+
+    \copybrief ANetAstrometryTask
+
+    Calculate an Exposure's zero-point given a set of flux measurements of stars matched to an input catalogue.
+    The type of flux to use is specified by ANetAstrometryConfig.fluxField.
```
TallJimbo: This looks like documentation for a photometric calibration Task.  That's not what this does, right?

python/lsst/meas/astrom/anetAstrometry.py
```diff
@@ -0,0 +1,464 @@
...
+        The reference catalog actually used is up to the implementation
+        of the solver; it will be manifested in the returned matches as
+        a list of lsst.afw.table.ReferenceMatch objects (\em i.e. of lsst.afw.table.Match with
+        \c first being of type lsst.afw.table.SimpleRecord and \c second type lsst.afw.table.SourceRecord ---
+        the reference object and matched object respectively).
+
+        \note
+        The input sources have the centroid slot moved to a new column "centroid.distorted"
+        which has the positions corrected for any known optical distortion;
+        the 'solver' (which is instantiated in the 'astrometry' member)
```
TallJimbo: Do you mean "astrometer" instead of "astrometry"?

python/lsst/meas/astrom/anetAstrometry.py
```diff
@@ -0,0 +1,464 @@
...
+        sourceCat.table.defineCentroid(self.distortedName)
+        try:
+            yield bbox # Execute 'with' block, providing bbox to 'as' variable
+        finally:
+            # Un-apply distortion
+            sourceCat.table.defineCentroid(oldCentroidName)
+            x0, y0 = exposure.getXY0()
+            wcs = exposure.getWcs()
+            if wcs:
+                wcs.shiftReferencePixel(-bbox.getMinX() + x0, -bbox.getMinY() + y0)
```
TallJimbo: It doesn't look like you're actually removing the distortion here in the `finally` block.

python/lsst/meas/astrom/anetAstrometry.py
```diff
@@ -0,0 +1,464 @@
...
+
+    @pipeBase.timeMethod
+    def astrometry(self, sourceCat, exposure, bbox=None):
```
TallJimbo: If the name of this method isn't being preserved for backwards-compatibility purposes, it should be changed.

python/lsst/meas/astrom/astrom.py
```diff
@@ -254,11 +259,11 @@ def useKnownWcs(self, sources, wcs=None, exposure=None, filterName=None, imageSi
         probably also want to turn OFF 'calculateSip'.
 
         This involves searching for reference sources within the WCS
-        area, and matching them to the given 'sources'.  If
+        area, and matching them to the given 'sourceCat'.  If
```
TallJimbo: I think it'd have been best to put all these trivial name changes in code you're not otherwise modifying in a separate commit.  I don't suppose it'd still be possible to separate them out?

python/lsst/meas/astrom/astrometry_net.i
```diff
@@ -243,64 +243,71 @@ void set_an_log(PTR(pexLog::Log) newlog);
...
-                  double ra, double dec, double radius,
-                  const char* idcol,
-                  std::vector<std::string> const& magnameVec,
-                  std::vector<std::string> const& magcolVec,
-                  std::vector<std::string> const& magerrcolVec,
-                  const char* stargalcol,
-                  const char* varcol,
-                  bool unique_ids=true)
+    /**
+    Find objects in a given region of the sky in an
```
TallJimbo: Docstring appears truncated.

python/lsst/meas/astrom/fitTanSipWcs.py
```diff
@@ -0,0 +1,57 @@
...
+class FitTanSipWcsConfig(pexConfig.Config):
+    order = pexConfig.RangeField(
+        doc = "order of SIP polynomial (0 for pure TAN WCS)",
+        dtype = int,
+        default = 5,
+        min = 0,
+    )
+
+class FitTanSipWcsTask(pipeBase.Task):
+    """!Match sources to position reference objects using optimistic pattern matching
```
TallJimbo: This looks like the docstring for a matcher, not a fitter.

python/lsst/meas/astrom/astrometry.py
r-owen: Good point. I copied the documentation and didn't change the tag.

include/lsst/meas/astrom/detail/utils.h
r-owen: I confess I am not very comfortable with abbreviated namespaces. I'll change it if you think it sufficiently important.

python/lsst/meas/astrom/astrometry.py
r-owen: I agree. This is pre-existing code. I'll try casting it and using 1 if the cast fails unless you think it's safe to just use it. That said, long is not likely for a frame number.

python/lsst/meas/astrom/astrometry.py
r-owen: I hate the old name, but was trying not to break too much stuff at once (it is a flag that is fairly widely used). Both of your choices make more sense and each has its charms.

python/lsst/meas/astrom/astrometry.py
r-owen: OK. I'll indent the showAstrometry calls as well.

include/lsst/meas/astrom/matchOptimisticB.h
r-owen: I decided it was a reasonable construct so I left it alone. It allows the code to easily work on pairs of any combination of ref objects and sources, which is something it needs to do.

But I did mean to make it internal to the .cc file.

include/lsst/meas/astrom/matchOptimisticB.h
r-owen: Yikes. I don't know how that snuck in there. I agree all on one line.

include/lsst/meas/astrom/matchOptimisticB.h
r-owen: Agreed.

python/lsst/meas/astrom/anetAstrometry.py
r-owen: Drat. Yes the old astrometry task had chunks of docs for the photometry task and I fixed it for the new astrometry task but forgot to fix it for the old one.

python/lsst/meas/astrom/anetAstrometry.py
r-owen: I don't think so. The member is called astrometry.

python/lsst/meas/astrom/anetAstrometry.py
r-owen: I'm afraid I didn't take the time to try to figure out the complications involved in distorting and undistorting pixels. This is old code that I only cosmetically touched. I'm hoping it will entirely go away, but if it sticks around it desperately needs an overhaul.

python/lsst/meas/astrom/anetAstrometry.py
r-owen: Again, I think the less touched in this task the better, until we truly overhaul it.

python/lsst/meas/astrom/astrom.py
r-owen: I doubt it. It took me several hours to squash my commits since I kept running into problems, but when I was done I discarded the old repo with 27 commits. Technically it's probably still on github but I really don't want to try that again. (As to sources: this is a pet peeve of mine; I find the name entirely unhelpful. It is a collection of some kind, but what kind? A catalog and a list have entirely different behavior. I felt a bit guilty switching, but I'd really like to see us switch everywhere. Names like sources make life very hard for newbies and people who don't touch this code often. Also collections named with a plural violate our coding standards, or at least did at one time.)

python/lsst/meas/astrom/astrometry_net.i
r-owen: Thanks. Good catch.

python/lsst/meas/astrom/fitTanSipWcs.py
r-owen: Good catch!

python/lsst/meas/astrom/loadReferenceObjects.py
```diff
@@ -0,0 +1,166 @@
...
+        """!Load reference objects that overlap a pixel-based rectangular region
+
+        The search algorith works by searching in a region in sky coordinates whose center is the center
+        of the bbox and radius is large enough to just include all 4 corners of the bbox.
+        Stars that lie outside the bbox are then trimmed from the list.
+
+        @param[in] bbox  bounding box for pixels (an lsst.afw.geom.Box2I or Box2D)
+        @param[in] wcs  WCS (an lsst.afw.image.Wcs)
+        @param[in] filterName  name of filter, or None or blank for the default filter
+        @param[in] calib  calibration, or None if unknown
```
TallJimbo: I don't see how you can support `calib` being `None` if the reference catalog is supposed to contain fluxes, not magnitudes.  And I'd actually thought that we'd settled on having magnitudes in the standard schema here, though I might be mistaken about that.

python/lsst/meas/astrom/matchOptimisticB.py
```diff
@@ -0,0 +1,241 @@
...
+        """!Remove stars from a catalog that are too near the center of M31 or M32
+        """
+        retStarCat = type(starCat)(starCat.table)
+        m31 = afwCoord.Coord(afwGeom.Point2D(10.681, 41.269))
+        m32 = afwCoord.Coord(afwGeom.Point2D(10.675, 40.864))
+        for star in starCat:
+            if (star.getCoord().angularSeparation(m31).asArcminutes() > 2.2 and
+                star.getCoord().angularSeparation(m32).asArcminutes() > 1.0):
+                retStarCat.append(star)
+        return retStarCat
```
TallJimbo: I suspect this code is carried over from some special-case processing of HSC M31 data, and I don't think it really belongs in a general-purpose matcher.  Or, rather, since it's clear we need to do this kind of rejection in some fields, it'd be better to make it somehow configurable rather than based on hard-coded positions.

python/lsst/meas/astrom/matchOptimisticB.py
```diff
@@ -0,0 +1,241 @@
...
+        if len(matchList) < minMatchedPairs:
+            self.log.warn("Number of matches is smaller than request")
+
+        return pipeBase.Struct(
+            matchList = matchList,
+        )
+
+    @pipeBase.timeMethod
+    def doMatch(self, refCat, sourceCat, wcs, refFluxField, numCleanSources, minMatchedPairs, verbose):
+        """!Match sources to position reference stars
```
TallJimbo: Name and description don't give any indication of what makes this different from `matchObjectsToSources`.  It seems like this might be a private method?

python/lsst/meas/astrom/matchOptimisticB.py
```diff
@@ -0,0 +1,241 @@
...
+        posRefBegInd = numCleanSources - numSources
+        configMatchDistPix = self.config.maxMatchDistArcSec/wcs.pixelScale().asArcseconds()
+
+        matchControl = MatchOptimisticBControl()
+        matchControl.refFluxField = refFluxField
+        matchControl.numBrightStars = self.config.numBrightStars
+        matchControl.minMatchedPairs = self.config.minMatchedPairs
+        matchControl.maxOffsetPix = self.config.maxOffsetPix
+        matchControl.numPointsForShape = self.config.numPointsForShape
+        matchControl.maxDeterminant = self.config.maxDeterminant
```
TallJimbo: Since you're using `LSST_CONTROL_FIELD` in C++, I'd expected you to wrap the control object into a Config object in Python.  Even though you have additional fields that aren't in the C++ control class, I think that'd be the way to go here, and it should let you get rid of the above assignment methods.

To do that, back in the definition of the config class, you'd use a decorator:
```
@pexConfig.wraps(MatchOptimisticBControl)
class MatchOptimisticBConfig(pexConfig.Config):
    # only define fields that aren't in the C++ control object
```
Then, here, you can just say:
```
matchControl = self.config.makeControl()
```

python/lsst/meas/astrom/utils.cc
```diff
@@ -53,282 +54,295 @@ read_column(fitstable_t* tag,
...
 
-    size_t const nMag = magcols.size();   /* number of magnitude[error] columns */
-    std::vector<mag_column_t>::const_iterator mc;
+    size_t const nMag = magColInfoList.size();   /* number of magnitude[error] columns */
+    std::vector<MagColInfo>::const_iterator mc;
 
-    for (mc = magcols.begin(); mc != magcols.end(); ++mc) {
-        if (mc->name.size() == 0) {
+    for (mc = magColInfoList.begin(); mc != magColInfoList.end(); ++mc) {
+        if (mc->filterName.size() == 0) {
```
TallJimbo: `mc->filterName.empty()` would be better.

And, while you can't use range-based `for` yet, you can use `auto` for the iterator type here.

python/lsst/meas/astrom/utils.cc
```diff
@@ -53,282 +54,295 @@ read_column(fitstable_t* tag,
...
         }
         // We enforce this condition because we convert the mags to fluxes, and
         // we need a flux to compute a flux error!
-        if (mc->magcol.size() == 0) {
-            throw LSST_EXCEPT(lsst::pex::exceptions::InvalidParameterError,
-                              "Magnitude column names cannot be empty string.");
+        if (mc->magCol.size() == 0) {
+        throw LSST_EXCEPT(lsst::pex::exceptions::InvalidParameterError,
+            "Magnitude column names cannot be empty strings.");
         }
```
TallJimbo: Indentation within this `if` block is off.

python/lsst/meas/astrom/utils.cc
```diff
@@ -53,282 +54,295 @@ read_column(fitstable_t* tag,
...
-             var = static_cast<bool*>(fitstable_read_column_inds(tag, varcol, boo, starinds, nstars));
-             if (!var) {
-                throw LSST_EXCEPT(lsst::pex::exceptions::NotFoundError,
-                                  str(boost::format("Unable to read data for %s from %s") %
-                                      varcol % ind->indexname));
+            if (id && uniqueIds) {
+                // remove duplicate IDs.
+
+                // FIXME -- this shouldn't be necessary once we get astrometry_net 0.40
+                // multi-index functionality in place.
```
TallJimbo: I know this isn't your change, but we should probably open an issue on this.

python/lsst/meas/astrom/utils.cc
```diff
@@ -53,282 +54,295 @@ read_column(fitstable_t* tag,
...
+                            // did not insert (this id has already been found);
+                            // drop this star.
+                        }
+                    }
+                    nstars = nkeep;
+                    // if they were all duplicate IDs...
+                    if (nstars == 0) {
+                        free(starinds);
+                        free(radecs);
+                        free(id);
```
TallJimbo: I know this isn't your code, but we *really* need to get rid of this C-style memory management, and replace it with smart pointers, especially considering that we have some exceptions being thrown in this code, and that could lead to memory leaks here.  Could you open an issue to deal with this in the future?

src/matchOptimisticB.cc
```diff
@@ -0,0 +1,672 @@
...
+        if (verbose) {
+            std::cout << "Catalog sizes: " << sourceSubCat.size() << " " << posRefSubCat.size() << std::endl;
+        }
+
+    /*    
+        CompareProxyFlux cmp = {posRefCat.getSchema().find<double>("flux").key};
+        std::sort(posRefProxyCat.begin(), posRefProxyCat.end(), cmp);
+        cmp = {sourceCat.getTable()->getPsfFluxKey()};
+        std::sort(sourceProxyCat.begin(), sourceProxyCat.end(), cmp);
+    */
```
TallJimbo: Would prefer to see this deleted or at least disabled with `#if 0` instead of comments.

tests/createWcsWithSip.py
```diff
@@ -71,7 +63,7 @@ def tearDown(self):
         import lsst.meas.astrom.astrometry_net as an
         an.finalize()
 
-    def testBigXy0(self):
+    def xtestBigXy0(self):
```
TallJimbo: Was disabling this test intentional?  If so, maybe use `@unittest.skip(msg)` instead?

tests/testFitTanSipWcsTask.py
```diff
@@ -0,0 +1,290 @@
...
+
+        if self.MatchClass == afwTable.ReferenceMatch:
+            refSchema = LoadReferenceObjectsTask.makeMinimalSchema(
+                filterNameList = ["r"], addFluxSigma=True, addIsPhotometric=True)
+            self.refCat = afwTable.SimpleCatalog(refSchema)
+        elif self.MatchClass == afwTable.SourceMatch:
+            refSchema = afwTable.SourceTable.makeMinimalSchema()
+            self.refCat = afwTable.SourceCatalog(refSchema)
+        else:
+            raise RuntimeError("Unsupport MatchClass=%r" % (self.MatchClass,))
```
TallJimbo: "Unsupported"?

tests/testLoadAstrometryNetObjects.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        loadANetObj = measAstrom.LoadAstrometryNetObjectsTask(config=self.config)
+
+        loadRes = loadANetObj.loadObjectsInBBox(bbox=self.bbox, wcs=self.wcs, filterName="r")
+        refCat = loadRes.refCat
+        self.assertEqual(loadRes.fluxField, "r_flux")
+        self.assertEqual(len(refCat), self.desNumStars)
+        self.assertObjInBBox(refCat=refCat, bbox=self.bbox, wcs=self.wcs)
+        schema = refCat.getSchema()
+        filterNameList = ['r'] # ['u', 'g', 'r', 'i', 'z']
```
TallJimbo: Did you mean to just use `r` here permanently?  Or was this intended to be a temporary change?

python/lsst/meas/astrom/matchOptimisticB.py
RobertLuptonTheGood: Is that cunning decorator documented somewhere?

ktlim: https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/pex_config.html under "Wrapping C++ Control Objects".

include/lsst/meas/astrom/detail/utils.h
PaulPrice: Then you should spell it `::lsst::`.

python/lsst/meas/astrom/anetAstrometry.py
PaulPrice: Have you removed the `astrometry.astrometer.solver` mess?

include/lsst/meas/astrom/detail/utils.h
r-owen: Sounds fine. And I'll get rid of the parens.

include/lsst/meas/astrom/detail/utils.h
r-owen: Why? The namespace is lsst::...

PaulPrice: Because `lsst::` would also match `::foo::bar::lsst::`

python/lsst/meas/astrom/astrometry.py
r-owen: I'm reluctant to change this, as it's a common pattern.

python/lsst/meas/astrom/anetAstrometry.py
r-owen: The new astrometry task has no such thing. I did not clean up the old a.net-based astrometry task, though it could use some cleanup.

python/lsst/meas/astrom/loadReferenceObjects.py
r-owen: I came to the opposite conclusion: it was easier and more natural to support calib being None if the reference loader continued to return flux instead of magnitudes. We should discuss this further.

python/lsst/meas/astrom/matchOptimisticB.py
r-owen: Good point. I do think this matcher needs to avoid those regions and I'm not sure it needs to avoid any others, but I intended to make it more general. I had a stab at it, but was not happy with the data types available in configs. I will take another look.

python/lsst/meas/astrom/matchOptimisticB.py
r-owen: Yes. I'll make it so and try to clarify the documentation.

python/lsst/meas/astrom/matchOptimisticB.py
r-owen: The problem is that MatchOptimisticBControl has some fields that should not be in MatchOptimisticBConfig and cannot be frozen. In a few cases there appears to be a 1:1 match, but somebody at HSC  decided iteration was needed, so the config value is taken as a parameter for the iteration.

python/lsst/meas/astrom/utils.cc
r-owen: Good change. This makes mc local to each for loop, which is a distinct improvement. I'll risk cbegin and hope gcc 4.4 can handle it -- buildbot will tell me (a web search suggested it could, but was not definitive)

python/lsst/meas/astrom/utils.cc
r-owen: Good idea. DM-2219. And with any luck we can just yank this code.

python/lsst/meas/astrom/utils.cc
r-owen: Yikes. I hadn't seen that. I agree.

src/matchOptimisticB.cc
r-owen: Deleted.

tests/createWcsWithSip.py
r-owen: Good catch. I suspect I was debugging and forgot to re-enable it.

tests/testLoadAstrometryNetObjects.py
r-owen: Good catch. I think I was just trying to speed things up while debugging.

python/lsst/meas/astrom/astrom.py
TallJimbo: On not separating the commits: fine.

On plural variable names vs. appending types: I generally disagree (I think including the types makes the code less readable and more fragile, as most containers are substitutable and we do sometimes substitute them).  But our coding standards are indeed on your side, and the most I'd advocate changing them in this respect would be to just let individual developers decide.

tests/testLoadAstrometryNetObjects.py
PaulPrice: Also, `['u', 'g', 'r', 'i', 'z']` is better spelt as `"ugriz"`.

python/lsst/meas/astrom/anetAstrometry.py
```diff
@@ -0,0 +1,453 @@
...
+            self.centroidYKey = schema.addField(self.distortedName + "_y", type="D",
+                                           doc="centroid distorted for astrometry solver")
+            self.centroidXErrKey = schema.addField(self.distortedName + "_xSigma", type="F",
+                                           doc="centroid distorted err for astrometry solver")
+            self.centroidYErrKey = schema.addField(self.distortedName + "_ySigma", type="F",
+                                           doc="centroid distorted err for astrometry solver")
+            self.centroidFlagKey = schema.addField(self.distortedName + "_flag", type="Flag",
+                                           doc="centroid distorted flag astrometry solver")
+            self.centroidKey = Point2DKey(self.centroidXKey, self.centroidYKey)
+            self.centroidErrKey = CovarianceMatrix2fKey((self.centroidXErrKey, self.centroidYErrKey))
```
TallJimbo: For future reference, the `FunctorKey` classes in afw and meas_base can help with adding all of these fields (and they can do it in a version-dependent way).  I don't think it's worth changing here, but I'll probably use that to simplify things when I sweep through after removing version 0 tables on DM-1766.

python/lsst/meas/astrom/matchOptimisticB.py
TallJimbo: Ah, sorry I never replied to this before: you can actually do this with the stuff in pex_config as well: just don't use `LSST_CONTROL_FIELD` to define the control class data members that shouldn't be converted into config fields (i.e. just make them regular public data members).  The `LSST_CONTROL_FIELD` macro's only purpose is to provide special static methods that the Python config-generator code can use to create the Python-side Field object, so it's a little confusing to see it used at all when that's not being used.

python/lsst/meas/photocal/PhotoCal.py
```diff
@@ -401,64 +414,68 @@ def extractMagArrays(self, matches, filterName, keys):
...
+                fluxErrKey = refSchema.find(fluxField + "Sigma").key
+                refFluxErrArr = np.array([m.first.get(fluxErrKey) for m in matches])
             except KeyError:
-                # Catalogue may not have flux uncertainties; HACK
+                # Reference catalogue may not have flux uncertainties; HACK
                 self.log.warn("Reference catalog does not have flux uncertainties for %s; using sqrt(flux)."
-                              % flux)
-                refFluxErr = np.sqrt(refFlux)
+                              % fluxField)
+                refFluxErrArr = np.sqrt(refFluxArr)
```
TallJimbo: I don't particularly like this default (and apparently neither did the original author, hence the "HACK").  Mind opening a new issue for this?

python/lsst/meas/photocal/PhotoCal.py
```diff
@@ -401,64 +414,68 @@ def extractMagArrays(self, matches, filterName, keys):
...
 
         # Fitting with error bars in both axes is hard, so transfer all
-        # the error to src, then convert to magnitude
-        fluxErr = np.hypot(srcFluxErr, refFluxErr)
-        magErr = fluxErr / (srcFlux * np.log(10))
+        # the error to src, then convert to magnitude;
+        # estimate the scaling first, to put ref errors in approximately the same system as src errors
+        meanSrcPerRef = afwMath.makeStatistics(srcFluxArr/refFluxArr, afwMath.MEANCLIP).getValue()
+        fluxErrArr = np.hypot(srcFluxErrArr, refFluxErrArr*meanSrcPerRef)
+        magErrArr = np.array([abMagErrFromFluxErr(fe, sf) for fe, sf in izip(fluxErrArr, srcFluxArr)])
```
TallJimbo: Your workaround here seems quite reasonable, but:
 - Please open a ticket to request a better solution.
 - I'm a little worried about whether your quick way of estimating the scaling can fail, and what happens if it does.

Given that the errors on the reference flux are likely to be approximately proportional to the errors on the source fluxes (because the same objects will be bright or faint in both catalogs), I wonder if an even less principled workaround of just dropping the reference flux errors would be more robust.  If we're typically taking the path I commented on above in which the reference flux error is set to just the square root of the flux (again, with no concern for the scale), that might actually be more principled.

python/lsst/meas/photocal/PhotoCal.py
r-owen: I have opened DM-2308 about this hack and the later hack of scaling reference flux errors. As Jim notes in the later comment it may be more robust to ignore reference flux error entirely, so I propose to adopt that solution until DM-2308 is properly dealt with.

python/lsst/meas/photocal/PhotoCal.py
r-owen: I have filed DM-2308 about this hack and the hack of using sqrt(refFlux) as the reference flux error if the error is unknown (though that is unlikely to be a problem for any reasonable photometric catalog).

Meanwhile I plan to adopt your suggestion of simply ignoring reference catalog errors until DM-2308 is dealt with. It is not ideal, but may be more robust than the current hacks.

python/lsst/meas/astrom/matchOptimisticB.py
r-owen: The problem I have with making some fields config fields and some not is that I would not have been able to predict in advance that some of the fields would not be fixed. The fact is that the C++ code is called repeatedly with various parameters tweaked in an effort to work around some problem that is never specified and for which I don�t have sufficient data to study it. It is maddening, as calling the C++ code repeatedly is almost certainly very, very inefficient. In short it is an ugly hack.

In any case I think this can all be cleaned up once we have a reasonable collection of data to test, to see in what ways the solver can go astray. Meanwhile I would rather leave things as they are, since the macros offer a nice place to put documentation and type information.

� Russell

P.S. github is being incredibly flakey today about letting me add inline notes, so I�ll try email.

tests/testLoadAstrometryNetObjects.py
r-owen: I respectfully disagree. Your suggestion certainly uses fewer characters but I feel it obscures what is going on and forces single-letter filter names.

python/lsst/meas/astrom/astrom.py
r-owen: For generic ordered containers I agree (and have continued to use �matches� instead of �matchList�), but I feel that tables act very differently than other ordered containers and some indication is in order, so I strongly prefer �sourceCat" to �sources�. To some extent this may be a C++ vs python thing; in C++ is often obvious what the type is by reading the header, but that is not true in Python. In trying to learn this code I have found many instances where the data types were not obvious and it really mattered.

python/lsst/meas/astrom/anetAstrometry.py
r-owen: I made as few changes to the existing a.net astrometry task as I could. In my opinion it needs major cleanup, including using the new standard for reference catalog schema and handling distortion in a simpler way (using a DistortedTanWcs). But I think that work should be left for another ticket.

I tried to remove all existing use of the Astrometry class except in the a.net solver, though I may have missed a few.

PR #4, opened 2015-03-16
meas_astrom Review for DM-420

TallJimbo:


TallJimbo:
Looks fine.  Either @r-owen or I will get rid of the remaining version 0 table use here on a future ticket.

PR #14, opened 2015-05-27
meas_astrom tickets/DM-2535: switch from compound fields to FunctorKeys

TallJimbo:


tests/testMatchOptimisticB.py
```diff
@@ -132,8 +133,9 @@ def computePosRefCatalog(self, sourceCat):
         refCat = afwTable.SimpleCatalog(minimalPosRefSchema)
         for source in sourceCat:
             refObj = refCat.addNew()
-            refObj.set("coord", source.get("coord"))
-            refObj.set("centroid", source.getCentroid())
+            refObj.setCoord(source.getCoord())
+            refObj.set("centroid_x", source.getX())
+            refObj.set("centroid_y", source.getY())
```
r-owen: It is practical to set the two fields at once using a centroidKey? If so, that would be cleaner.

TallJimbo: I disagree that it's cleaner.  It's a lot of setup for a one-liner, and while it'd be more efficient if it was done repeatedly in production code, I don't think it matters in test code.

r-owen: Point taken. I will note that tests are often used as examples, and using keys is a better example. But as you say...it's a lot of extra code for no performance gain.

TallJimbo:
Merged to master on tickets/DM-1766.

PR #19, opened 2015-09-17
tickets/DM-3824

r-owen:


tests/createWcsWithSip.py
```diff
@@ -79,6 +78,7 @@ def testBigXy0(self):
             src.set(yKey, src.get(yKey) - 500 + y0)
         bbox = afwGeom.Box2I(afwGeom.Point2I(x0, y0), afwGeom.Extent2I(1000, 1000))
         res = self.astrom.determineWcs2(cat, bbox=bbox)
+        self.assertTrue(res.sipWcs is not None, "Failed to fit SIP terms")
```
timj:
```python
self.assertIsNotNone(res.sipWcs, "Failed to fit SIP terms")
```

tests/testMakeMatchStatistics.py
```diff
@@ -0,0 +1,165 @@
...
+    def tearDown(self):
+        del self.wcs
+        del self.bboxD
+        del self.sourceCentroidKey
+        del self.sourceCat
+        del self.refCoordKey
+        del self.refCat
+        del self.matchList
+
+    def xtestMakeMatchStatistics(self):
```
timj: Why is this test disabled?

r-owen: That was an accident. Thank you for catching it.

tests/createWcsWithSip.py
r-owen: Is the error message sufficiently clearer to justify that change? I am keen on assertLess(a, b) instead of assertTrue(a<b) and the like because they give better error messages and are easier to read. I'm not sure I see the value in this one.

timj: The message will also be clearer since it will tell you that `res.sipWcs` was not none as opposed to telling you that False is not True. I can't think of a reason why we wouldn't want to go for the more explicit test.

r-owen: The message includes the code that was False, so the message is sufficiently clear. With conditionals, e.g. a > b you don't see the individual values. I'm willing to do it, but I don't think it's very useful and those ugly long method names start to get to me after awhile. I didn't even know that one existed. It feels like clutter to me.

timj: Thanks for the prodding. I've had a look in more detail and in the case where no message is included the explicit test is clearer than `assertTrue`. In the case where a contextual message is included both options result in the same output. I think I'd still prefer the `assertIsNotNone` version if only because it makes it easier for me to find the old-style `assertTrue` code that still exists.

jdswinbank:
This work is long since complete.

PR #22, opened 2016-02-22
patch meas_astrom v11.0 to prevent a.d.n SIGSEVs

jhoblitt:


timj:
If this patch is by @mjuric how come the commit is not by him? Use `git commit --author`. And yes, the commit needs a description and a typo fix in the commit message.

jhoblitt:
I lifted the patch out of an unrelated repo, and it is part of a large changeset.  It wouldn't have been helpful to export the entire changeset as a diff or play changes with sub-tree in order to cherry-pick it.  If you want, I can fudge the commit author...

timj:
I would prefer author attribution was done properly even for small manually applied patches.

mjuric:
FWIW, here's the original patch (and a description of what's going on -- lessons learned all around!): https://github.com/mjuric/conda-lsst/commit/c38cc6b3f8dca9726d9307d455569277fa438487

jhoblitt:
@mjuric Thanks -- I never would have found that due to a rebase.  I will fix up this PR in a few mins.

RobertLuptonTheGood:
-O3 #defining NDEBUG is new.  And bad.

mjuric:
@RobertLuptonTheGood Yup. Bad @dstndstn :). Re-enabling asserts may be a good follow-up to this patch...

Offending line: https://github.com/dstndstn/astrometry.net/blob/master/util/makefile.common#L251

dstndstn:
You don't have debug builds?  Tsk thyself

dstndstn:
And the underlying bug was where?  I didn't see any patches flowing upstream to astrometry.net

RobertLuptonTheGood:
With C++ building -O0 is something like 10x slower and optimising shouldn't generate bugs.  Traditionally building NDEBUG and non-debugged are orthogonal issues (and the former is only done after profiling).

mjuric:
> And the underlying bug was where? I didn't see any patches flowing upstream to astrometry.net

Fair point. I never figured out what the underlying bug was :(.

I did e-mail some Dustin guy to check that it was a bug, though ;-).

PR #23, opened 2016-03-07
DM-4692: Refactor ProcessCcdTask and sub-tasks

PaulPrice:


PaulPrice:
The SNR changes need to be squashed together.

PR #24, opened 2016-03-19
DM-5532

r-owen:


python/lsst/meas/astrom/catalogStarSelector.py
```diff
@@ -76,25 +66,72 @@ def __call__(self, source):
...
+    - psf flux >= config.fluxLim
+    - psf flux <= config.fluxMax (not checked if fluxMax == 0)
+
+    @section meas_astrom_catalogStarSelector_Initialize  Task initialisation
+
+    @copydoc \_\_init\_\_
+
+    @section meas_astrom_catalogStarSelector_IO  Invoking the Task
+
+    Like all star selectors, the main method is `run`.
```
TallJimbo: Should call out the fact that this is an unusual `StarSelector` for which `usesMatches=True`, and hence that argument is required.

jhoblitt:
Can one of the admins verify this patch?

python/lsst/meas/astrom/catalogStarSelector.py
r-owen: Good idea

PR #27, opened 2016-05-09
Change from boost::math to std

pschella:


TallJimbo:
Aside from `<cmath>` duplication, include order was correct before.

PR #28, opened 2016-05-10
ANetAstrometryConfig: add alias refObjLoader

SimonKrughoff:
ANetAstrometryConfig names the 'refObjLoader' as 'solver', but users
(specifically, when getting the match metadata) can refer to it as
'refObjLoader'. This alias makes that work and allows us to use
astrometry.net again.

### DIFF
--- a/python/lsst/meas/astrom/anetAstrometry.py
+++ b/python/lsst/meas/astrom/anetAstrometry.py
@@ -48,6 +48,11 @@ class ANetAstrometryConfig(pexConfig.Config):
     rejectIter = pexConfig.RangeField(dtype=int, default=3, doc="Rejection iterations for Wcs fitting",
                                       min=0)
 
+    @property
+    def refObjLoader(self):
+        """An alias, for a uniform interface with the standard AstrometryTask"""
+        return self.solver
+
 
     ## \addtogroup LSST_task_documentation
     ## \{


python/lsst/meas/astrom/anetAstrometry.py
```diff
@@ -48,6 +48,11 @@ class ANetAstrometryConfig(pexConfig.Config):
     rejectIter = pexConfig.RangeField(dtype=int, default=3, doc="Rejection iterations for Wcs fitting",
                                       min=0)
 
+    @property
+    def refObjLoader(self):
+        """An alias, for a uniform interface with the standard AstrometryTask"""
+        return self.solver
+
```
SimonKrughoff: I'm confused why we haven't seen this before.  Is it because no one uses this class, or did you need to do something special?

PaulPrice: The change triggering the problem is fairly recent (about a month ago; DM-5686), and I suspect no-one has used our astrometry.net solver since that went in.

r-owen: Having looked at the code, I agree it's reasonable for now. This change can be reverted when we overhaul the way reference loading is done (e.g. pass in a reference loader to the the astrometry solver in order to support the new index loader).

PR #29, opened 2016-05-24
Replace boost::scoped_array with std::unique_ptr

jdswinbank:


src/matchOptimisticB.cc
```diff
@@ -4,8 +4,8 @@
 #include <iostream>
 #include <utility>
 #include <vector>
+#include <memory>
```
jdswinbank: Should be alphabetically sorted ([coding standard 4-10](http://developer.lsst.io/en/latest/coding/cpp_style_guide.html#include-statements-should-be-sorted-and-grouped)).

PR #33, opened 2016-06-21
Tickets/dm 6620

parejkoj:
Make `LoadAstrometryNetObjectsTask.__init__()` create a default config object if it isn't passed one.


parejkoj:
I'm not checking the autopep8 stuff, but the other change looks good.

PR #36, opened 2016-07-22
Tickets/dm 6352

r-owen:


SimonKrughoff:
It looks like the sdssrefcat exists in both meas_astrom and pipe_tasks.  It seems like that is unnecessary.  Can't it just live in meas_astrom?

r-owen:
That data appears in both packages, and is identical in both. I'm not thrilled about this duplication, but in general we prefer test data in packages to be self-contained, so that modifying test data in one package won't mysteriously break tests in other packages. And indeed both packages had duplicate a.net versions of this data (tests/astrometry_net_data/photocal). I was able to delete one of those, and the other will move to the new a.net wrapper package as part of separating moving the a.net wrapper code into its own package.

PR #37, opened 2016-08-02
DM-6984: Remove unused pex_logging dependency

hchiang2:


hchiang2:
No longer needed because it was done in DM-6999 (PR#38)

PR #38, opened 2016-08-05
DM-6999: Logging framework migration

hchiang2:


tests/openFiles.py
```diff
@@ -87,7 +87,7 @@ def getAstrom(self):
 
         conf = measAstrom.ANetBasicAstrometryConfig()
         return measAstrom.ANetBasicAstrometryTask(config=conf, andConfig=andconfig,)
-        #logLevel=pexLog.Log.DEBUG)
+        #logLevel=pexLog.Log.TRACE)
```
isullivan: I suggest you just delete this dead code, rather than updating it.

tests/testJoinMatchListWithCatalog.py
```diff
@@ -54,14 +54,14 @@ def setUp(self):
         mi.assign(smallExposure.getMaskedImage(), smallExposure.getBBox())
 
         logLevel = Log.INFO
-        #logLevel = Log.DEBUG
+        #logLevel = Log.TRACE
```
isullivan: As above, I suggest you just delete this commented-out code.

isullivan:
I had only very minor comments on the code. I don't have the new lsst.log so the changed tests all fail for me, but assuming that they do all pass with the new framework this branch can be merged.

tests/openFiles.py
hchiang2: It's not removed.

tests/testJoinMatchListWithCatalog.py
hchiang2: I'm leaving this one un-touched.

PR #39, opened 2016-08-09
Tickets/dm 7117

r-owen:


python/lsst/meas/astrom/match.py
```diff
@@ -0,0 +1,102 @@
...
+    def calculateCircle(self, catalog):
+        """!Calculate a circle enclosing the catalog
+
+        @param catalog[in]  Catalog we will encircle (lsst.afw.table.SourceCatalog)
+        @return Struct with center (lsst.afw.coord.Coord) and radius (lsst.afw.geom.Angle)
+        """
+        coordList = [src.getCoord() for src in catalog]
+        center = afwCoord.averageCoord(coordList)
+        radius = max(center.angularSeparation(coord) for coord in coordList)
+        return Struct(center=center, radius=radius)
```
r-owen: Is this algorithm good enough for the intended purpose? It is unlikely to be the smallest circle that encloses the catalog (imagine a cluster of many points and a few outliers off to one side).

python/lsst/meas/astrom/match.py
```diff
@@ -0,0 +1,102 @@
...
+     - @ref meas_astrom_match_Purpose
+     - @ref meas_astrom_match_Initialize
+     - @ref meas_astrom_match_IO
+     - @ref meas_astrom_match_Config
+     - @ref meas_astrom_match_Example
+
+    @section meas_astrom_match_Purpose  Description
+
+    Match sources to reference objects. This is often done as a preliminary step to fitting an astrometric
+    or photometric solution. For details about the matching algorithm see match.h
```
r-owen: Please say a bit more about how this differs from our usual matcher, e.g. that it performs no scaling or rotation. Otherwise I worry that users will get confused about what this does vs the other matcher(s).

The reference to match.h is a bit opaque (and the case is wrong). I suggest something like "The matching algorithm is matchRaDec, which is defined in lsst/afw/table/Match.h".

python/lsst/meas/astrom/createMatchMetadata.py
```diff
@@ -4,7 +4,25 @@
...
+        @param[in] radius: Minimum radius for selecting sources (lsst.afw.geom.Angle)
+        @param[in] filterName: Name of filter (str) or None
+        """
+        PropertyList.__init__(self)
+        ctrCoord = ctrCoord.toIcrs()
+        self.add('RA', ctrCoord.getRa().asDegrees(), 'field center in degrees')
+        self.add('DEC', ctrCoord.getDec().asDegrees(), 'field center in degrees')
+        self.add('RADIUS', radius.asDegrees(), 'field radius in degrees, minimum')
+        self.add('SMATCHV', 1, 'SourceMatchVector version number')
+        self.add('FILTER', filterName or "UNKNOWN", 'filter name for photometric data')
```
r-owen: Looking at the old code...are we likely to get a name of "_unknown_" if the filter is unknown, and if so, should it be changed to "UNKNOWN"?

python/lsst/meas/astrom/match.py
PaulPrice: I think this is the best we can do given the available API for the `refObjLoader`.  It may not be the most efficient algorithm but I don't think it's far off if the sources in the `catalog` are constrained to be in the same area, like they are if they're all from the same image; and that's my use case.  If not, the user may have to write and `retarget` a more efficient implementation.

python/lsst/meas/astrom/createMatchMetadata.py
PaulPrice: I think you're referring to the line in astrometry.py:

        if filterName == "_unknown_":
            filterName = None

I have no idea where the `"_unknown_"` comes from, and I've never seen it in practise.  I think you added that code; do you know what it's for?

tests/testMatch.py
```diff
@@ -0,0 +1,97 @@
...
+class MatchTestCase(lsst.utils.tests.TestCase):
+    """Tests for lsst.meas.astrom.MatchTask"""
+
+    def setUp(self):
+        self.butler = Butler(RefCatDir)
+        refObjLoader = LoadIndexedReferenceObjectsTask(butler=self.butler)
+        center = lsst.afw.coord.IcrsCoord(215.5*lsst.afw.geom.degrees, 53.0*lsst.afw.geom.degrees)
+        radius = 0.5*lsst.afw.geom.degrees
+        self.filter = "r"
+        self.references = refObjLoader.loadSkyCircle(center, radius, self.filter).refCat
```
r-owen: I suggest using the standard name "refCat" as it's less confusing than renaming variables.

tests/testMatch.py
```diff
@@ -0,0 +1,97 @@
...
+
+from lsst.utils import getPackageDir
+from lsst.daf.persistence import Butler
+from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask
+
+RefCatDir = os.path.join(getPackageDir("meas_astrom"), "tests", "data", "sdssrefcat")
+
+numpy.random.seed(12345)
+
+
```
r-owen: Please add this standard bit of boilerplate, as per https://developer.lsst.io/coding/python_testing.html 
"Memory and file descriptor leak testing" section

    def setup_module(module):
        lsst.utils.tests.init()

PR #42, opened 2016-09-14
python3 port Tickets/dm 7288

parejkoj:


examples/getSourceSet.py
```diff
@@ -353,8 +355,8 @@ def writeSourceSet(sourceSet, outfile="-"):
         fd = open(outfile, "w")
 
     for s in sourceSet:
-        print >> fd, s.getId(), s.getXAstrom(), s.getYAstrom(), s.getRa(
-        ), s.getDec(), s.getPsfFlux(), s.getFlagForDetection()
+        print(s.getId(), s.getXAstrom(), s.getYAstrom(), s.getRa(
+        ), s.getDec(), s.getPsfFlux(), s.getFlagForDetection(), file=fd)
```
timj: to my eyes, the line break here is weird. I'd expect the new line to occur at `s.getRa(),`

python/lsst/meas/astrom/astrometryNetDataConfig.py
```diff
@@ -3,6 +3,9 @@
 mechanism, but we need nested lists, so we do this home-brew version
 instead.
 '''
+from past.builtins import execfile
```
timj: we should make a note to fix this in a python3 way at some point.

tests/testMakeMatchStatistics.py
```diff
@@ -102,7 +103,7 @@ def testMakeMatchStatistics(self):
         """
         np.random.seed(47)
         distList = list((np.random.random_sample([self.numMatches]) - 0.5) * 10)
-        for dist, match in itertools.izip(distList, self.matchList):
+        for dist, match in zip(distList, self.matchList):
```
timj: `from builtins import zip` at top.

tests/testOpenFiles.py
```diff
@@ -68,9 +71,11 @@ class OpenFilesTest(unittest.TestCase):
 
     def setUp(self):
         self.originalLimits = resource.getrlimit(resource.RLIMIT_NOFILE)
-        print 'NOFILE rlimit:', self.originalLimits
-        resource.setrlimit(resource.RLIMIT_NOFILE, (10, self.originalLimits[1]))
-        print 'NOFILE rlimit:', resource.getrlimit(resource.RLIMIT_NOFILE)
+        # don't want this to depend on how many files e.g. py.test has open at the start.
+        currentOpenFiles = len(getOpenFiles())
+        print('NOFILE rlimit:', self.originalLimits)
+        resource.setrlimit(resource.RLIMIT_NOFILE, (currentOpenFiles+5, self.originalLimits[1]))
```
timj: spaces need around `+` operator.

examples/getSourceSet.py
parejkoj: That's fair.

python/lsst/meas/astrom/astrometryNetDataConfig.py
parejkoj: Fix what? I would have thought all the past and builtins imports would be fixed together when we go to python3 only?

examples/getSourceSet.py
timj: and the indenting looks wrong.

tests/testMakeMatchStatistics.py
parejkoj: Done.

tests/testOpenFiles.py
parejkoj: Done, though my linter doesn't catch that one.

python/lsst/meas/astrom/astrometryNetDataConfig.py
timj: We have generally used `past.builtins` to give us compatibility with behavioral changes between 2 and 3. In particular using `long` and `basestring` as they are providing semantics to python 2 that do not exist on 3. `execfile` is not like that and it just a function that needs to be replaced with a modern approach.

tests/testOpenFiles.py
timj: that's because we disable "spaces around operator" warnings because it's an area where there is vagueness in PEP8 and where we disagree with some interpretations.

timj:
Please remember to add `python_future` to the table file.

PR #46, opened 2016-10-03
meas_astrom/DM-7736: Avoid signed/unsigned integer comparison warnings.

TallJimbo:


src/matchOptimisticB.cc
```diff
@@ -535,7 +535,7 @@ namespace astrom {
         if (posRefBegInd < 0) {
             throw LSST_EXCEPT(pexExcept::InvalidParameterError, "posRefBegInd < 0");
         }
-        if (posRefBegInd >= posRefCat.size()) {
+        if (static_cast<size_t>(posRefBegInd) >= posRefCat.size()) {
```
kfindeisen: It's an interface change, but we might change `posRefBegInd` to `size_t`. It's clear from both the name and the validation block that this is supposed to be an index in the normal C++ sense.

PR #47, opened 2016-10-03
meas_astrom/DM-3549: new SIP fitter

TallJimbo:
This adds a new fitter for SIP distortions in the form of a number of C++ classes (several polynomial-based transform classes and a fitter) and a new Python task that is intended as a drop-in replacement for `FitTanSipWcsTask`.

Some caveats:

 - I have not yet enabled the new task, `FitSipDistortionTask`, as the default; I think it needs a bit more testing before we do that.  I do however expect it to fully replace `FitTanSipWcsTask` eventually, so I have not tried to more strictly define the interface for WCS fitters or add a registry as we do with other pluggable tasks.  I also have ignored some (minor) code duplication between the tasks (especially their config classes) that would be better to clean up with common base class if we expected them to coexist for a long time.

 - The new transform classes do not inherit from `XYTransform` or any other common base class, and that probably limits their usefulness to this particular application.  I think we'll want to refactor that when we have our new WCS/transform library, but doing it before then is probably premature.

### DIFF
--- a/include/lsst/meas/astrom/PolynomialTransform.h
+++ b/include/lsst/meas/astrom/PolynomialTransform.h
@@ -1,0 +1,257 @@
+// -*- LSST-C++ -*-
+
+/*
+ * LSST Data Management System
+ * Copyright 2016 LSST/AURA
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+#ifndef LSST_MEAS_ASTROM_PolynomialTransform_INCLUDED
+#define LSST_MEAS_ASTROM_PolynomialTransform_INCLUDED
+
+#include "ndarray/eigen.h"
+#include "lsst/afw/geom/AffineTransform.h"
+
+namespace lsst { namespace meas { namespace astrom {
+
+class SipForwardTransform;
+class SipReverseTransform;
+class ScaledPolynomialTransform;
+
+/**
+ *  A 2-d coordinate transform represented by a pair of standard polynomials
+ *  (one for each coordinate).
+ *
+ *  PolynomialTransform instances should be confined to a single thread.
+ */
+class PolynomialTransform {
+public:
+
+    /**
+     *  Convert a ScaledPolynomialTransform to an equivalent PolynomialTransform.
+     */
+    static PolynomialTransform convert(ScaledPolynomialTransform const & other);
+
+    /**
+     *  Convert a SipForwardTransform to an equivalent PolynomialTransform.
+     */
+    static PolynomialTransform convert(SipForwardTransform const & other);
+
+    /**
+     *  Convert a SipReverseTransform to an equivalent PolynomialTransform.
+     */
+    static PolynomialTransform convert(SipReverseTransform const & other);
+
+    /**
+     *  Construct a new transform from existing coefficient arrays.
+     *
+     *  For both input arguments, the array element at [p, q] corresponds
+     *  to the polynomial term x^p y^q.
+     *
+     *  Both arrays are expected be square and triangular; if N is the
+     *  order of the transform, both arrays should be (N+1)x(N+1), and
+     *  elements with p + q > N should be zero.
+     */
+    PolynomialTransform(
+        ndarray::Array<double const,2,2> const & xCoeffs,
+        ndarray::Array<double const,2,2> const & yCoeffs
+    );
+
+    /**
+     *  Copy constructor.
+     *
+     *  Coefficient arrays are deep-copied.
+     */
+    PolynomialTransform(PolynomialTransform const & other);
+
+    /**
+     *  Move constructor.
+     *
+     *  Coefficient arrays are moved.
+     */
+    PolynomialTransform(PolynomialTransform && other);
+
+    /**
+     *  Copy assignment.
+     *
+     *  Coefficient arrays are deep-copied.
+     */
+    PolynomialTransform & operator=(PolynomialTransform const & other);
+
+    /**
+     *  Move constructor.
+     *
+     *  Coefficient arrays are moved.
+     */
+    PolynomialTransform & operator=(PolynomialTransform && other);
+
+    /// Lightweight swap.
+    void swap(PolynomialTransform & other);
+
+    /// Return the order of the polynomials.
+    int getOrder() const { return _xCoeffs.rows() - 1; }
+
+    /**
+     * 2-D polynomial coefficients that compute the output x coordinate.
+     *
+     * Indexing the result by [p][q] gives the coefficient of
+     * @f$x_{\mathrm{in}}^p\,y_{\mathrm{in}}^q@f$.
+     */
+    ndarray::Array<double const,2,2> getXCoeffs() const { return _xCoeffs.shallow(); }
+
+    /**
+     * 2-D polynomial coefficients that compute the output x coordinate.
+     *
+     * Indexing the result by [p][q] gives the coefficient of
+     * @f$x_{\mathrm{in}}^p\,y_{\mathrm{in}}^q@f$.
+     */
+    ndarray::Array<double const,2,2> getYCoeffs() const { return _yCoeffs.shallow(); }
+
+    /**
+     * Return an approximate affine transform at the given point.
+     */
+    afw::geom::AffineTransform linearize(afw::geom::Point2D const & in) const;
+
+    /**
+     * Apply the transform to a point.
+     */
+    afw::geom::Point2D operator()(afw::geom::Point2D const & in) const;
+
+private:
+
+    PolynomialTransform(int order);
+
+    friend PolynomialTransform compose(afw::geom::AffineTransform const & t1, PolynomialTransform const & t2);
+    friend PolynomialTransform compose(PolynomialTransform const & t1, afw::geom::AffineTransform const & t2);
+    friend class ScaledPolynomialTransformFitter;
+    friend class SipForwardTransform;
+    friend class SipReverseTransform;
+    friend class ScaledPolynomialTransform;
+
+    ndarray::EigenView<double,2,2> _xCoeffs;
+    ndarray::EigenView<double,2,2> _yCoeffs;
+    mutable Eigen::VectorXd _u; // workspace for operator() and linearize
+    mutable Eigen::VectorXd _v;
+};
+
+/**
+ *  A 2-d coordinate transform represented by a lazy composition of an AffineTransform,
+ *  a PolynomialTransform, and another AffineTransform.
+ *
+ *  ScaledPolynomialTransform instances should be confined to a single thread.
+ */
+class ScaledPolynomialTransform {
+public:
+
+    /**
+     *  Convert a PolynomialTransform to an equivalent ScaledPolynomialTransform.
+     *
+     *  This simply inserts identity AffineTransforms before and after applying
+     *  the given PolynomialTransform.
+     */
+    static ScaledPolynomialTransform convert(PolynomialTransform const & poly);
+
+    /**
+     *  Convert a SipForwardTransform to an equivalent ScaledPolynomialTransform.
+     *
+     *  The input transform's CRPIX offset and CD matrix scaling are used to define
+     *  the input and output affine transforms, respectively, leaving the polynomial
+     *  coefficients unmodified.
+     */
+    static ScaledPolynomialTransform convert(SipForwardTransform const & sipForward);
+
+    /**
+     *  Convert a SipForwardTransform to an equivalent ScaledPolynomialTransform.
+     *
+     *  The input transform's CD matrix scaling and CRPIX offset are used to define
+     *  the input and output affine transforms, respectively, leaving the polynomial
+     *  coefficients unmodified.
+     */
+    static ScaledPolynomialTransform convert(SipReverseTransform const & sipReverse);
+
+    /**
+     *  Construct a new ScaledPolynomialTransform from its constituents.
+     *
+     *  @param[in]  poly           A PolynomialTransform to be applied to points.
+     *                             after the inputScaling transform.
+     *  @param[in]  inputScaling   An AffineTransform to be applied immediately
+     *                             to input points.
+     *  @param[in]  outputScalingInverse  An AffineTransform to be applied to points
+     *                                    after the PolynomialTransform.
+     */
+    ScaledPolynomialTransform(
+        PolynomialTransform const & poly,
+        afw::geom::AffineTransform const & inputScaling,
+        afw::geom::AffineTransform const & outputScalingInverse
+    );
+
+    ScaledPolynomialTransform(ScaledPolynomialTransform const & other) = default;
+
+    ScaledPolynomialTransform(ScaledPolynomialTransform && other) = default;
+
+    ScaledPolynomialTransform & operator=(ScaledPolynomialTransform const & other) = default;
+
+    ScaledPolynomialTransform & operator=(ScaledPolynomialTransform && other) = default;
+
+    void swap(ScaledPolynomialTransform & other);
+
+    /// Return the polynomial transform applied after the input scaling.
+    PolynomialTransform const & getPoly() const { return _poly; }
+
+    /// Return the first affine transform applied to input points.
+    afw::geom::AffineTransform const & getInputScaling() const { return _inputScaling; }
+
+    /// Return the affine transform applied to points after the polynomial transform.
+    afw::geom::AffineTransform const & getOutputScalingInverse() const { return _outputScalingInverse; }
+
+    /**
+     * Return an approximate affine transform at the given point.
+     */
+    afw::geom::AffineTransform linearize(afw::geom::Point2D const & in) const;
+
+    /**
+     * Apply the transform to a point.
+     */
+    afw::geom::Point2D operator()(afw::geom::Point2D const & in) const;
+
+private:
+    friend class ScaledPolynomialTransformFitter;
+    PolynomialTransform _poly;
+    afw::geom::AffineTransform _inputScaling;
+    afw::geom::AffineTransform _outputScalingInverse;
+};
+
+/**
+ *  Return a PolynomialTransform that is equivalent to the composition t1(t2())
+ *
+ *  The returned composition would be exact in ideal arithmetic, but may suffer from
+ *  significant round-off error for high-order polynomials.
+ */
+PolynomialTransform compose(afw::geom::AffineTransform const & t1, PolynomialTransform const & t2);
+
+/**
+ *  Return a PolynomialTransform that is equivalent to the composition t1(t2())
+ *
+ *  The returned composition would be exact in ideal arithmetic, but may suffer from
+ *  significant round-off error for high-order polynomials.
+ */
+PolynomialTransform compose(PolynomialTransform const & t1, afw::geom::AffineTransform const & t2);
+
+}}} // namespace lsst::meas::astrom
+
+#endif // !LSST_MEAS_ASTROM_PolynomialTransform_INCLUDED
--- a/include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
+++ b/include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
@@ -135,7 +135,7 @@ public:
      *   - initial_{x,y}:      reference positions transformed by initialWcs.
      *   - model_{x,y}:        reference positions transformed by current
      *                         best-fit distortion.
-     *   - valid:              Flag that is set for objects that have not been
+     *   - rejected            Flag that is set for objects that have been
      *                         rejected as outliers.
      */
     static ScaledPolynomialTransformFitter fromMatches(
--- a/include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
+++ b/include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
@@ -1,0 +1,319 @@
+// -*- LSST-C++ -*-
+
+/*
+ * LSST Data Management System
+ * Copyright 2016 LSST/AURA
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+#ifndef LSST_MEAS_ASTROM_TanSipFitter_INCLUDED
+#define LSST_MEAS_ASTROM_TanSipFitter_INCLUDED
+
+#include "lsst/pex/config.h"
+#include "lsst/afw/geom/LinearTransform.h"
+#include "lsst/afw/image/TanWcs.h"
+#include "lsst/afw/table/Catalog.h"
+#include "lsst/afw/table/BaseRecord.h"
+#include "lsst/meas/astrom/PolynomialTransform.h"
+#include "lsst/meas/astrom/SipTransform.h"
+
+namespace lsst { namespace meas { namespace astrom {
+
+
+/**
+ *  Control object for outlier rejection in ScaledPolynomialTransformFitter.
+ *
+ *  See ScaledPolynomialTransformFitter::rejectOutliers for more information.
+ */
+class OutlierRejectionControl {
+public:
+
+    OutlierRejectionControl() :
+        nSigma(3), nClipMin(1), nClipMax(5)
+    {}
+
+    LSST_CONTROL_FIELD(nSigma, double, "Number of sigma to clip at.");
+
+    LSST_CONTROL_FIELD(nClipMin, int, "Always clip at least this many matches.");
+
+    LSST_CONTROL_FIELD(nClipMax, int, "Never clip more than this many matches.");
+
+};
+
+
+/**
+ *  A fitter class for scaled polynomial transforms
+ *
+ *  This class allows for iteration between actual fitting, outlier rejection,
+ *  and estimation of intrinsic scatter.  It also provides access to the
+ *  current model for debugging via an afw::table::BaseCatalog that contains
+ *  the input values, output values, and the best-fit-transformed input values.
+ *
+ *  ScaledPolynomialTransformFitter has two public construction methods:
+ *
+ *   - fromMatches fits a transform that maps intermediate world coordinates
+ *     to pixel coordinates, using as inputs a match of reference coordinates
+ *     to measured source positions.  This sets up the fitter to do outlier
+ *     rejection and intrinsic scatter estimation.
+ *
+ *   - fromGrid fits an inverse to an existing transform, using the
+ *     to-be-inverted transform to populate a grid of positions.  Because the
+ *     to-be-inverted transform is assumed to be known exactly, fitters
+ *     initialized with this method have no need for outlier rejection or
+ *     scatter estimation.
+ *
+ *  In either case, the fitter creates affine transforms that map the input
+ *  and output data points onto [-1, 1].  It then fits a polynomial
+ *  transform that, when composed with the input scaling transform and the
+ *  inverse of the output scaling transform, maps the input data points to the
+ *  output data points.
+ *
+ *  The fitter can be used in an outlier-rejection loop with the following
+ *  pattern (with user-defined convergence criteria):
+ *  @code
+ *  while (true) {
+ *      fitter.fit();
+ *      if (converged) break;
+ *      fitter.updateModel();
+ *      fitter.computeIntrinsicScatter();
+ *      fitter.rejectOutliers();
+ *  }
+ *  @endcode
+ *  This pattern fits a model, uses that model to transform the input
+ *  points, estimates intrinsic scatter from the differences between
+ *  model-transformed points and output data points, and finally rejects
+ *  outliers by sigma-clipping those differences before repeating.
+ *
+ *  ScaledPolynomialTransformFitter instances should be confined to a single thread.
+ */
+class ScaledPolynomialTransformFitter {
+public:
+
+    /**
+     *  Initialize a fit from intermediate world coordinates to pixels using
+     *  source/reference matches.
+     *
+     *  @param[in] maxOrder         Maximum polynomial order for the fit.
+     *
+     *  @param[in] matches          Vector of source-reference matches.
+     *                              The centroids and centroid errors from the
+     *                              sources are used, along with the coords
+     *                              from the reference objects.
+     *
+     *  @param[in] initialWcs       Initial WCS, used to transform reference
+     *                              positions to intermediate world coordinates and
+     *                              populate the "ref" field in the data catalog.
+     *
+     *  @param[in] intrinsicScatter Initial intrinsic scatter to be added in
+     *                              quadrature with source measurement errors in
+     *                              setting the uncertainty on each match.
+     *
+     *  This initializes the data catalog with the following fields:
+     *   - ref_id:             reference object ID
+     *   - src_id:             source ID from the match
+     *   - src_{x,y}:          source position in pixels
+     *   - src_{xSigma,ySigma,x_y_Cov}:  uncertainty on source positions,
+     *                         including measurement errors and the inferred
+     *                         intrinsic scatter.
+     *   - intermediate_{x,y}: reference positions in intermediate world
+     *                         coordinates.
+     *   - initial_{x,y}:      reference positions transformed by initialWcs.
+     *   - model_{x,y}:        reference positions transformed by current
+     *                         best-fit distortion.
+     *   - valid:              Flag that is set for objects that have not been
+     *                         rejected as outliers.
+     */
+    static ScaledPolynomialTransformFitter fromMatches(
+        int maxOrder,
+        afw::table::ReferenceMatchVector const & matches,
+        afw::image::Wcs const & initialWcs,
+        double intrinsicScatter
+    );
+
+    /**
+     *  Initialize a fit that inverts an existing transform by evaluating and
+     *  fitting to points on a grid.
+     *
+     *  @param[in] maxOrder   Maximum polynomial order for the fit.
+     *
+     *  @param[in] bbox       Bounding box for the grid in the input
+     *                        coordinates of the toInvert transform (or
+     *                        equivalently the output coordinates of the
+     *                        transform to be fit).
+     *
+     *  @param[in] nGridX     Number of grid points in the X direction.
+     *
+     *  @param[in] nGridY     Number of grid points in the Y direction.
+     *
+     *  @param[in] toInvert   Transform to invert
+     *
+     *  This initializes the data catalog with the following fields:
+     *   - output:  grid positions passed as input to the toInvert transform,
+     *              treated as output data points when fitting its inverse.
+     *   - input:   grid positions generated as the output of the toInvert
+     *              transform, treated as input data points when fitting its
+     *              inverse.
+     *   - model:   best-fit-transformed input points.
+     *
+     *  The updateIntrinsicScatter and rejectOutliers methods cannot be used
+     *  on fitters initialized using this method.
+     */
+    static ScaledPolynomialTransformFitter fromGrid(
+        int maxOrder,
+        afw::geom::Box2D const & bbox,
+        int nGridX, int nGridY,
+        ScaledPolynomialTransform const & toInvert
+    );
+
+    /**
+     *  Perform a linear least-squares fit of the polynomial coefficients.
+     *
+     *  @param[in]  order    The maximum order of the polynomial transform.
+     *                       If negative (the default) the maxOrder from
+     *                       construction is used.
+     *
+     *  After fitting, updateModel() should be called to apply the model
+     *  transform to the input points if the user wants to make use of the
+     *  data catalog for diagnostics.  Calling fit() alone is sufficient if
+     *  the user is only interested in getting the model transform itself.
+     */
+    void fit(int order=-1);
+
+    /**
+     *  Update the 'model' field in the data catalog using the current best-
+     *  fit transform.
+     *
+     *  Should be called after fit() and before updateIntrinsicScatter() if
+     *  the fitter is being used in an outlier-rejection loop.
+     */
+    void updateModel();
+
+    /**
+     *  Infer the intrinsic scatter in the offset between the data points and
+     *  the best-fit model, and update the uncertainties accordingly.
+     *
+     *  The scatter is calculated as the RMS scatter that maximizes the
+     *  likelihood of the current positional offsets (assumed fixed) assuming
+     *  the per-data-point uncertainties are the quadrature sum of the
+     *  intrinsic scatter and that source's centroid measurement uncertainty.
+     *
+     *  Should be called after updateModel() and before rejectOutliers() if
+     *  the fitter is being used in an outlier-rejection loop.
+     */
+    double updateIntrinsicScatter();
+
+    /**
+     *  Return the current estimate of the intrinsic scatter.
+     *
+     *  Because the intrinsic scatter is included in the uncertainty used in
+     *  both the fit and outlier rejection, this may be called either before
+     *  updateIntrinsicScatter() to obtain the scatter used in the last such
+     *  operation or after updateIntrinsicScatter() to obtain the scatter to
+     *  be used in the next operation.
+     */
+    double getIntrinsicScatter() const { return _intrinsicScatter; }
+
+    /**
+     *  Mark outliers in the data catalog using sigma clipping.
+     *
+     *  A data point is declare an outlier if:
+     *   - the offset between the model prediction and the data position,
+     *     weighted by the uncertainty (including inferred intrinsic scatter)
+     *     of that point exceeds ctrl.nSigma
+     *   - the number of points already with larger weighted offsets is less
+     *     than ctrl.nClipMax.
+     *  In addition, the worst (in weighted offset) ctrl.nClipMin data points
+     *  are always rejected.
+     *
+     *  @return A pair of the smallest rejected weighted offset (units of sigma)
+     *          and the number of point rejected.
+     *
+     *  Should be called after updateIntrinsicScatter() and before fit() if
+     *  the fitter is being used in an outlier rejection loop.
+     */
+    std::pair<double,size_t> rejectOutliers(OutlierRejectionControl const & ctrl);
+
+    /**
+     *  Return a catalog of data points and model values for diagnostic purposes.
+     *
+     *  The values in the returned catalog should not be modified by the user.
+     *
+     *  For information about the schema, either introspect it programmatically
+     *  or see fromMatches and fromGrid.
+     */
+    afw::table::BaseCatalog const & getData() const { return _data; }
+
+    /**
+     *  Return the best-fit transform
+     *
+     *  If f is an instance of this class, then for an
+     *  arbitrary point p:
+     *  @code
+     *  f.getTransform()(p) == f.getOutputScaling().invert()(f.getPoly()(f.getInputScaling()(p)))
+     *  @endcode
+     */
+    ScaledPolynomialTransform const & getTransform() const { return _transform; }
+
+    /**
+     *  Return the polynomial part of the best-fit transform.
+     */
+    PolynomialTransform const & getPoly() const { return _transform.getPoly(); }
+
+    /**
+     *  Return the input scaling transform that maps input data points to [-1, 1].
+     */
+    afw::geom::AffineTransform const & getInputScaling() const { return _transform.getInputScaling(); }
+
+    /**
+     *  Return the output scaling transform that maps output data points to [-1, 1].
+     */
+    afw::geom::AffineTransform const & getOutputScaling() const { return _outputScaling; }
+
+private:
+
+    class Keys;
+
+    ScaledPolynomialTransformFitter(
+        afw::table::BaseCatalog const & data,
+        Keys const & keys,
+        int maxOrder,
+        double intrinsicScatter,
+        afw::geom::AffineTransform const & inputScaling,
+        afw::geom::AffineTransform const & outputScaling
+    );
+
+    double computeIntrinsicScatter() const;
+
+    // Normally it's not safe to use a reference as a data member because the
+    // class holding it can't control when the referenced object gets
+    // destroyed, but this points to one of two singletons (which never get
+    // destroyed).
+    Keys const & _keys;
+    double _intrinsicScatter;
+    afw::table::BaseCatalog _data;
+    afw::geom::AffineTransform _outputScaling;
+    ScaledPolynomialTransform _transform;
+    // 2-d generalization of the Vandermonde matrix: evaluates polynomial at
+    // all data points when multiplied by a vector of packed polynomial
+    // coefficients.
+    Eigen::MatrixXd _vandermonde;
+};
+
+}}} // namespace lsst::meas::astrom
+
+#endif // !LSST_MEAS_ASTROM_TanSipFitter_INCLUDEDtr
--- a/include/lsst/meas/astrom/SipTransform.h
+++ b/include/lsst/meas/astrom/SipTransform.h
@@ -1,0 +1,372 @@
+// -*- LSST-C++ -*-
+
+/*
+ * LSST Data Management System
+ * Copyright 2016 LSST/AURA
+ *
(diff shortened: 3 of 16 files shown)


include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+ *
+ */
+class SipReverseTransform : public SipTransformBase {
+public:
+
+    /**
+     *  Convert a PolynomialTransform to an equivalent SipReverseTransform.
+     *
+     *  @param[in]   poly           PolynomialTransform to convert.
+     *  @param[in]   pixelOrigin    CRPIX @f$(u_0,v_0)@f$
```
r-owen: Again, anywhere you mention CRPIX please be explicit about the zero convention (that's safer than putting it in one place and hoping folks stumble across it).

include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+    afw::geom::Point2D operator()(afw::geom::Point2D const & xy) const;
+
+private:
+    friend class PolynomialTransform;
+    friend class ScaledPolynomialTransform;
+    afw::geom::LinearTransform _cdInverse;
+};
+
+/**
+ *  Create a new TanWcs from a pair of SIP transforms and the sky origin.
```
r-owen: I suggest you say "TAN SIP WCS" instead of a "TanWcs"; the latter is a poorly named class that should go away; the former is a more general concept that should remain after the WC rewrite.

python/lsst/meas/astrom/astromLib.i
```diff
@@ -11,26 +11,61 @@ Python interface to lsst::meas::astrom
...
 %import "lsst/afw/table/tableLib.i"
 %import "lsst/afw/image/wcs.i"
 %import "lsst/afw/math/statistics.i"
+%import "lsst/afw/geom/geomLib.i"
 
 %include "lsst/meas/astrom/matchOptimisticB.h"
 
+%include "std_pair.i"
+
+namespace std { // have to do this in namespace block so Swig can find size_t
```
r-owen: weird. You can't just use std::size_t?

python/lsst/meas/astrom/astromLib.i
```diff
@@ -11,26 +11,61 @@ Python interface to lsst::meas::astrom
...
 %shared_ptr(lsst::meas::astrom::MatchOptimisticBControl);
 
 %import "lsst/afw/table/tableLib.i"
 %import "lsst/afw/image/wcs.i"
 %import "lsst/afw/math/statistics.i"
+%import "lsst/afw/geom/geomLib.i"
 
 %include "lsst/meas/astrom/matchOptimisticB.h"
 
+%include "std_pair.i"
```
r-owen: If this could safely go higher (e.g. right near `%include "ndarray.i"` I think it would make the code a bit easier to read; we usually put standard and 3rd party includes together near the top when we can get away with it.

Another option worth considering is moving the new code into a separate .i file. There is a fair amount of it, so it may enhance readability.

python/lsst/meas/astrom/astromLib.i
```diff
@@ -11,26 +11,61 @@ Python interface to lsst::meas::astrom
...
+
+namespace std { // have to do this in namespace block so Swig can find size_t
+%template(DoubleSizePair) pair<double,size_t>;
+}
+
+%declareNumPyConverters(ndarray::Array<double const,2,2>)
```
r-owen: Here's another line I suggest move way up if that's safe to do.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+            Reference object catalog, or None.
+            If provided then all centroids are updated with the new WCS,
+            otherwise only the centroids for ref objects in matches are updated.
+            Required fields are "centroid_x", "centroid_y", "coord_ra", and "coord_dec".
+        sourceCat : :cpp:class:`lsst::afw::table::SourceCatalog`
+            Source catalog, or None.
+            If provided then coords are updated with the new WCS;
+            otherwise only the coords for sources in matches are updated.
+            Required fields are "slot_Centroid_x", "slot_Centroid_y",
+            "coord_ra", and "coord_dec".
```
r-owen: Don't you also need centroid uncertainty?

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+                rx, ry = record.get(refKey)
+                disp.dot("x", rx, ry, size=10, ctype=colors[0])
+                mx, my = record.get(modelKey)
+                disp.dot("o", mx, my, size=10, ctype=colors[0])
+                disp.line([(rx, ry), (mx, my)], ctype=colors[0])
+                sx, sy = record.get(srcKey)
+                sErr = record.get(srcErrKey)
+                sEllipse = lsst.afw.geom.ellipses.Quadrupole(sErr[0, 0], sErr[1, 1], sErr[0, 1])
+                disp.dot(sEllipse, sx, sy, ctype=colors[1])
+        print("Dropping into debugger to allow inspection of display. Type 'continue' when done.")
```
r-owen: Why not just display each iteration in a new frame without stopping? It seems less disruptive.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def makeInitialWcs(self, matches, wcs):
+        """Generate a guess Wcs from the astrometric matches
+
+        We create a Wcs anchored at the center of the matches, with the scale
+        of the input Wcs.  This is necessary because matching returns only
+        matches with no estimated Wcs, and the input Wcs is a wild guess.
+        We're using the best of each: positions from the matches, and scale
+        from the input Wcs.
+
```
r-owen: This really surprises me. The initial WCS ought to include an estimate of optical distortion, and so ought to be pretty decent. Still...I assume you found you needed to do this instead of trusting the initial WCS. I just wish I knew why.

include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+ */
+#ifndef LSST_MEAS_ASTROM_SipTransform_INCLUDED
+#define LSST_MEAS_ASTROM_SipTransform_INCLUDED
+
+#include "lsst/afw/geom/LinearTransform.h"
+#include "lsst/meas/astrom/PolynomialTransform.h"
+
+
+namespace lsst {
+
```
r-owen: Better yet, put all `namespace foo { namespace bar { ...` on the same line, as with the other files

include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+    SipTransformBase(SipTransformBase const & other) = default;
+    SipTransformBase(SipTransformBase && other) = default;
+    SipTransformBase & operator=(SipTransformBase const & other) = default;
+    SipTransformBase & operator=(SipTransformBase && other) = default;
+
+    void swap(SipTransformBase & other) {
+        std::swap(_pixelOrigin, other._pixelOrigin);
+        std::swap(_cdMatrix, other._cdMatrix);
+        _poly.swap(other._poly);
+    }
```
pschella: Any particular reason you are not using ADL for `swap`?

include/lsst/meas/astrom/PolynomialTransform.h
```diff
@@ -0,0 +1,253 @@
...
+    friend PolynomialTransform compose(PolynomialTransform const & t1, afw::geom::AffineTransform const & t2);
+    friend class ScaledPolynomialTransformFitter;
+    friend class SipForwardTransform;
+    friend class SipReverseTransform;
+    friend class ScaledPolynomialTransform;
+
+    ndarray::EigenView<double,2,2> _xCoeffs;
+    ndarray::EigenView<double,2,2> _yCoeffs;
+    mutable Eigen::VectorXd _u; // workspace for operator() and linearize
+    mutable Eigen::VectorXd _v;
```
pschella: Thread safe?

src/ScaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,465 @@
...
+            }
+        }
+    }
+}
+
+void ScaledPolynomialTransformFitter::fit(int order) {
+    if (order < 0) {
+        order = _transform.getPoly().getOrder();
+    }
+    int const packedSize = detail::computePackedSize(order);
```
r-owen: If order is greater than the max order computed I think it will read off the end of an array. I suggest a test here with a useful exception that includes the max allowed and the value provided. That means you always must call `_transform.getPoly().getOrder()` which suggests setting a new variable maxOrder to that, then modifying the test above to use maxOrder. That would clarify the `if (order < 0)` test a bit, as well.

src/ScaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,465 @@
...
+    Eigen::VectorXd vx = Eigen::VectorXd::Zero(nGood);
+    Eigen::VectorXd vy = Eigen::VectorXd::Zero(nGood);
+    // sxx, syy, sxy: (2x2) blocks of the covariance matrix S, each of which is
+    // individually diagonal.
+    Eigen::ArrayXd sxx(nGood);
+    Eigen::ArrayXd syy(nGood);
+    Eigen::ArrayXd sxy(nGood);
+    Eigen::Matrix2d outS = _outputScaling.getLinear().getMatrix();
+    for (std::size_t i1 = 0, i2 = 0; i1 < _data.size(); ++i1) {
+        if (!_keys.valid.isValid() || _data[i1].get(_keys.valid)) {
```
r-owen: What is `_keys.valid`? `valid.isValid()` is a bit mind-bending Please document the Keys argument of the constructor (and all the other arguments) as requested in an earlier comment. Without that I have no idea what `_keys` is and what fields it contains.

Also, is `_keys.valid.isValid()` efficient to call? If in doubt, it might be better to set a local const variable and use that in the loop.

src/ScaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,465 @@
...
+
+#include "boost/math/tools/minima.hpp"
+
+#include "lsst/meas/astrom/ScaledPolynomialTransformFitter.h"
+#include "lsst/meas/astrom/detail/polynomialUtils.h"
+#include "lsst/afw/table/aggregates.h"
+#include "lsst/afw/table/Match.h"
+#include "lsst/afw/math/LeastSquares.h"
+
+#define LSST_ScaledPolynomialTransformFitter_TEST_IN_PLACE 1
```
r-owen: Please document what this does. Do you really want it left true for production code?

src/ScaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,465 @@
...
+            }
+            ++i2;
+        }
+    }
+    // Do a blockwise inverse of S.  Note that the result F is still symmetric
+    Eigen::ArrayXd fxx = 1.0/(sxx - sxy.square()/syy);
+    Eigen::ArrayXd fyy = 1.0/(syy - sxy.square()/sxx);
+    Eigen::ArrayXd fxy = -(sxy/sxx)*fyy;
+#ifdef LSST_ScaledPolynomialTransformFitter_TEST_IN_PLACE
+    assert((sxx*fxx + sxy*fxy).isApproxToConstant(1.0));
```
r-owen: I suspect we will disagree about this, but this section is presently on by default and I'm very nervous about asserts being run in production code. I'd be much happier with raising an exception so the user gets a more useful message and other code has a chance to continue if the fitter fails.

src/ScaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,465 @@
...
+            maxMeasurementVariance = std::max(maxMeasurementVariance, ca2);
+            ++nGood;
+        }
+    }
+    directVariance /= nGood;
+
+    // Function that computes the -log likelihood of the current deltas with
+    // the variance modeled as described above.
+    auto logLikelihood = [this](double intrinsicVariance) {
+        // Uncertainties in the table right now include the old intrinsic scatter; need to
```
r-owen: Is it important to use this syntax instead of defining a named function (perhaps in the anonymous namespace)? I think the latter would be easier to read.

src/ScaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,465 @@
...
+    );
+    return std::sqrt(result.first);  // return RMS instead of variance
+}
+
+
+std::pair<double,std::size_t> ScaledPolynomialTransformFitter::rejectOutliers(
+    OutlierRejectionControl const & ctrl
+) {
+    if (!_keys.valid.isValid()) {
+        throw LSST_EXCEPT(
```
r-owen: I already admitted to not knowing what `_keys.valid` is (or `_keys`) but this test really surprises me. The very fact that the `_keys.valid.isValid()` is true tells you that you are trying to reject outliers? Why? A comment here might be helpful.

src/SipTransform.cc
```diff
@@ -0,0 +1,185 @@
...
+
+PTR(afw::image::TanWcs) makeWcs(
+    SipForwardTransform const & sipForward,
+    SipReverseTransform const & sipReverse,
+    afw::coord::Coord const & skyOrigin
+) {
+    if (!sipForward.getPixelOrigin().asEigen().isApprox(sipReverse.getPixelOrigin().asEigen())) {
+        throw LSST_EXCEPT(
+            pex::exceptions::InvalidParameterError,
+            "SIP forward and reverse transforms have inconsistent CRPIX"
```
r-owen: Print the values?

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+import unittest
+import numpy
```
r-owen: Please change to `import numpy as np`

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+    origin = lsst.afw.geom.Point2D(*numpy.random.randn(2))
+    cd = lsst.afw.geom.LinearTransform(numpy.random.randn(2,2))
+    poly = makeRandomPolynomialTransform(order, sip=False)
+    return SipReverseTransform(origin, cd, poly)
+
+
+class TransformTestMixin(object):
+
+    def checkTransformEquivalence(self, a, b, rtol=1E-8):
+        for i in xrange(10):
```
r-owen: Please use `range` and `from builtins import range` for python 3 compatibility. More generally, I suggest a futurize stage 2 pass on this code to make sure it all should be python 3 compatible.

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+
+def makeRandomSipReverseTransform(order):
+    origin = lsst.afw.geom.Point2D(*numpy.random.randn(2))
+    cd = lsst.afw.geom.LinearTransform(numpy.random.randn(2,2))
+    poly = makeRandomPolynomialTransform(order, sip=False)
+    return SipReverseTransform(origin, cd, poly)
+
+
+class TransformTestMixin(object):
+
```
r-owen: To improve the readability I suggest adding a `makeRandom` method that raises `NotImlpementedError` or using `abc` to make this an abstract base class, and documenting `makeRandom` (the API and the need to override it).

include/lsst/meas/astrom/detail/polynomialUtils.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#ifndef LSST_MEAS_ASTROM_DETAIL_polynomialUtils_h_INCLUDED
+#define LSST_MEAS_ASTROM_DETAIL_polynomialUtils_h_INCLUDED
+
+#include "Eigen/Core"
+
+namespace lsst { namespace meas { namespace astrom { namespace detail {
+
+inline int computePackedOffset(int order) {
```
TallJimbo: Done (locally; will commit and push after aggregating minor comments).

include/lsst/meas/astrom/detail/polynomialUtils.h
```diff
@@ -0,0 +1,64 @@
...
+    return (order*(order+1))/2;
```
TallJimbo: Done.

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+        """
+        order = 3
+        xc = makeRandomCoefficientMatrix(order + 1)
+        yc = makeRandomCoefficientMatrix(order + 1)
+        p = PolynomialTransform(xc, yc)
+        self.assertEqual(p.getOrder(), order)
+        self.assertClose(p.getXCoeffs(), xc, atol=0, rtol=0)
+        self.assertClose(p.getYCoeffs(), yc, atol=0, rtol=0)
+        # Test that the coefficients are not a view.
+        xc[0, 0] = 50.0
```
r-owen: It is not self-evident that the existing value cannot be 50. Please consider incrementing the existing value (after saving the initial value to use in the comparison test).

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+        """
+        poly = makeRandomPolynomialTransform(4)
+        affine = lsst.afw.geom.AffineTransform(
+            lsst.afw.geom.LinearTransform(numpy.random.randn(2,2)),
+            lsst.afw.geom.Extent2D(*numpy.random.randn(2))
+        )
+        composed1 = lsst.meas.astrom.compose(poly, affine)
+        composed2 = lsst.meas.astrom.compose(affine, poly)
+        self.checkTransformEquivalence(composed1, lambda p: poly(affine(p)))
+        self.checkTransformEquivalence(composed2, lambda p: affine(poly(p)))
```
r-owen: The test doesn't quite seem to do what is advertised: check that the result of composing gives the expected result. I suggest you define a python transform that applies affine and then poly (or vice-versa) and check equivalence (similar to what `ScaledPolynomialTransformTestCase .testConstruction` does below)

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
```diff
@@ -0,0 +1,275 @@
...
+    /**
+     *  Initialize a fit that inverts an existing transform by evaluating and
+     *  fitting to points on a grid.
+     *
+     *  @param[in] maxOrder         Maximum polynomial order for the fit.
+     *
+     *  @param[in] bbox             Bounding box for the grid in the input
+     *                              coordinates of the toInvert transform
+     *                              (or equivalently the output coordinates
+     *                              of the transform to be fit).
```
TallJimbo: In the way it's used in `FitSipDistortionTask`, the bounding box has to be in intermediate world coordinates, and given that we only start with the bounding box of the image and we're trying to *find* the mapping from pixel coordinates to intermediate world coordinates here, our only choice is to transform the image bounding box using (part of) the initial guess WCS.

 I don't think it's appropriate to include that description in the docs for `fromGrid`, since it's more general than that, but I will add code comments to `FitSipDistortionTask` where I'm computing the bbox.

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+            fwd, rev,
+            lsst.afw.coord.IcrsCoord(crval, lsst.afw.geom.degrees)
+        )
+        def t1(p):
+            sky = lsst.afw.coord.IcrsCoord(crval + lsst.afw.geom.Extent2D(p), lsst.afw.geom.degrees)
+            return rev(wcs.skyToIntermediateWorldCoord(sky))
+        def t2(p):
+            sky = lsst.afw.coord.IcrsCoord(crval + lsst.afw.geom.Extent2D(p), lsst.afw.geom.degrees)
+            return wcs.skyToPixel(sky)
+        self.checkTransformEquivalence(t1, t2)
```
r-owen: Is it worth also checking fwd in the same way, with another pair of functions?

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+        makeRandomPolynomialTransform(order, sip=True)
+    )
+
+def makeRandomSipReverseTransform(order):
+    origin = lsst.afw.geom.Point2D(*numpy.random.randn(2))
+    cd = lsst.afw.geom.LinearTransform(numpy.random.randn(2,2))
+    poly = makeRandomPolynomialTransform(order, sip=False)
+    return SipReverseTransform(origin, cd, poly)
+
+
```
r-owen: This is a very nice set of tests. However, I feel as if too much is contained in one file. I suggest you move some of this to a library module, such as `lsst.meas.astrom.testUtils`, including `TransformTestMixin`. Then divide the remaining subclasses of `TransformTestMixin` into individual test files with more descriptive names.

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
```diff
@@ -0,0 +1,275 @@
...
+    static ScaledPolynomialTransformFitter fromGrid(
+        int maxOrder,
+        afw::geom::Box2D const & bbox,
+        int nGridX, int nGridY,
+        ScaledPolynomialTransform const & toInvert
+    );
+
+    /**
+     *  Perform a linear least-squares fit of the polynomial coefficients.
+     *
```
TallJimbo: Done.  I've added a bit to each of these functions and a code block to the class docs describing the typical iteration pattern.

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
```diff
@@ -0,0 +1,275 @@
...
+     *  assuming the per-data-point uncertainties are the quadrature sum
+     *  of the intrinsic scatter and that source's centroid measurement
+     *  uncertainty.
+     */
+    double updateIntrinsicScatter();
+
+    /**
+     *  Return the current estimate of the intrinsic scatter.
+     */
+    double getIntrinsicScatter() const { return _intrinsicScatter; }
```
TallJimbo: The intrinsic scatter is actually included in the uncertainty on the data points, so one might want to call this before `updateIntrinsicScatter` to obtain the scatter used in the last call to `fit` or `rejectOutliers`.  Calling it after `updateIntrinsicScatter` tells you the scatter that will be used in the next call to `fit` or `rejectOutliers`.  (I've added that to the docs for `getIntrinsicScatter`).

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
```diff
@@ -0,0 +1,275 @@
...
+     *
+     *  For information about the schema, either introspect it programmatically
+     *  or see fromMatches and fromGrid.
+     */
+    afw::table::BaseCatalog const & getData() const { return _data; }
+
+    /**
+     *  Return the best-fit transform
+     *
+     *  If f is a ScaledPolynomialTransformFitter instance, then for an
```
TallJimbo: Done.

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+class TransformTestMixin(object):
+
+    def checkTransformEquivalence(self, a, b, rtol=1E-8):
```
r-owen: Please rename to `assert...`, perhaps `assertTransformsNearlyEqual`.

Please also consider moving this to `afw.geom.testUtils`, such that it becomes a standard part of `lsst.utils.tests.TestCase` with a name that reflects the fact that the transforms it tests are 2d (e.g. `assert2DTransformsNearlyEqual`) and with a user-specified domain.

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
```diff
@@ -0,0 +1,275 @@
...
+        Keys const & keys,
+        int maxOrder,
+        double intrinsicScatter,
+        afw::geom::AffineTransform const & inputScaling,
+        afw::geom::AffineTransform const & outputScaling
+    );
+
+    double computeIntrinsicScatter() const;
+
+    Keys const & _keys;  // safe as member because this points to one of two singletons.
```
TallJimbo: Updated to:
```
    // Normally it's not safe to use a reference as a data member because the
    // class holding it can't control when the referenced object gets
    // destroyed, but this points to one of two singletons (which never get
    // destroyed).
```

tests/testTransforms.py
```diff
@@ -0,0 +1,468 @@
...
+
+
+class ScaledPolynomialTransformFitterTestCase(lsst.utils.tests.TestCase):
+
+    def setUp(self):
+        numpy.random.seed(50)
+
+    def testFromMatches(self):
+        # Setup artifical matches that correspond to a known (random) PolynomialTransform.
+        order = 3
```
r-owen: I suggest moving this code into a new method, e.g. `makeMatchList`, to make the test easier to read

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
```diff
@@ -0,0 +1,275 @@
...
+    double _intrinsicScatter;
+    afw::table::BaseCatalog _data;
+    afw::geom::AffineTransform _outputScaling;
+    ScaledPolynomialTransform _transform;
+    Eigen::MatrixXd _vandermonde;
```
TallJimbo: Done; looking it up (https://en.wikipedia.org/wiki/Vandermonde_matrix), I actually noticed that it's a 2-d generalization of the true Vandermonde matrix (which is defined for 1-d polynomials), so it'll be good to clarify that even in the case where the reader does recognize the name.

include/lsst/meas/astrom/SipTransform.h
TallJimbo: Fine.  I couldn't find any way of formatting the opening and partial closing and reopening of the namespaces that I liked, but your proposal is as good as any.

include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+ *
+ *  This class simply provides some getters for its derived classes.
+ *  It should not be used directly, and does not define a polymorphic
+ *  interface.
+ */
+class SipTransformBase {
+public:
+
+    /**
+     *  Return the pixel origin (CRPIX) of the transform.
```
TallJimbo: It's all zero-based.  I'll try to find all the places where I use the term.

tests/testTransforms.py
r-owen: Also, this test has a number of minor flake8 linter errors. Please run flake8 on this and all the other new python code you added and touched.

include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+    afw::geom::LinearTransform const & getCDMatrix() const { return _cdMatrix; }
+
+    /**
+     *  Return the polynomial component of the transform (A,B) or (AP,BP).
+     */
+    PolynomialTransform const & getPoly() const { return _poly; }
+
+protected:
+
+    SipTransformBase(
```
TallJimbo: "class->constructor", you mean?  Done.

include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+ *   - @f$(u_0,v_0)@f$ is the pixel origin @f$\mathrm{CRVAL}@f$.
+ *   - @f$\mathrm{A}@f$, @f$\mathrm{B}@f$ are the polynomial coefficients of
+ *     the forward transform.
+ *
+ *  The SIP convention encourages (but does not require) nulling the zeroth-
+ *  and first-order elements of
+ *  @f$\mathrm{A}@f$ and @f$\mathrm{B}@f$, which ensures the representation of
+ *  a given transform is unique.  This also makes fitting a SIP transform to
+ *  data a nonlinear operation, as well as making the conversion from standard
+ *  polynomial transforms to SIP form impossible in general.
```
TallJimbo: Done.

include/lsst/meas/astrom/SipTransform.h
```diff
@@ -0,0 +1,357 @@
...
+        afw::geom::Point2D const & pixelOrigin,
+        afw::geom::LinearTransform const & cdMatrix
+    );
+
+    /**
+     *  Convert a ScaledPolynomialTransform to an equivalent SipForwardTransform.
+     *
+     *  @param[in]   scaled         ScaledPolynomialTransform to convert.
+     *  @param[in]   pixelOrigin    CRPIX @f$(u_0,v_0)@f$
+     *  @param[in]   cdMatrix       CD matrix @f$Z@f$
```
TallJimbo: I did at one point, but that was before a lot of copy-and-pasting, and this is a good reminder to do a final check before I merge.

include/lsst/meas/astrom/SipTransform.h
TallJimbo: Done.

python/lsst/meas/astrom/astromLib.i
TallJimbo: Correct.  That was what I tried first.  I think I could have also fixed it by not using `std::` in the C++ header file, but I'm not sure that's valid C++ even though my compiler allows it.

python/lsst/meas/astrom/astromLib.i
TallJimbo: I was actually planning to put this in the custom .i file for the transform classes, since it's only needed by them.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+        dtype=int,
+        default=0
+    )
+    nClipMax = lsst.pex.config.Field(
+        doc="Maximum number of matches to reject when sigma-clipping",
+        dtype=int,
+        default=1
+    )
+    maxScatterArcsec = lsst.pex.config.RangeField(
+        doc="maximum median scatter of a WCS fit beyond which the fit fails (arcsec); " +
```
TallJimbo: Done (also capitalized "order" and "number" for the same reason).

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+        dtype=int,
+        default=100,
+    )
+    nGridY = lsst.pex.config.Field(
+        doc="Number of Y grid points used to invert the SIP reverse transform.",
+        dtype=int,
+        default=100,
+    )
+    gridBorder = lsst.pex.config.Field(
+        doc=("When setting the gird region, how much to extend the image "
```
TallJimbo: Done.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+        default=50.0,
+    )
+
+
+class FitSipDistortionTask(lsst.pipe.base.Task):
+    """Fit a TAN-SIP WCS given a list of reference object/source matches
+
+    FitSipDistortionTask is a drop-in replacement for
+    :py:class:`lsst.meas.astrom.FitTanSinWcsTask`.  It is built on fundamentally
+    stronger fitting algorithms, but has received significantly less testing.
```
TallJimbo: [DM-7934](https://jira.lsstcorp.org/browse/DM-7934)

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+       config.(...).astometry.wcsFitter.retarget(FitSipDistortionTask)
+
+    Algorithm
+    ---------
+
+    The algorithm used by FitSipDistortionTask involves three steps:
+
+     - We set the CRVAL and CRPIX reference points to the mean position, while
+       holding the CD matrix fixed to its input value.  This work is done by
+       the makeInitialWcs method.
```
TallJimbo: I thought this was a summary, though I see that I just said "mean position" when I should have said "mean position of all the matches".  Does that clarify things?

I think we could add a config option to leave CRVAL and/or CRPIX fixed at the positions from the given initial WCS, but this matches the behavior of `FitTanSipWcsTask` and I didn't see anything wrong with that.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+            - match.first (reference object) centroid,
+            - match.second (source) centroid
+            - match.distance (on sky separation, in radians)
+        initWcs : :cpp:class:`lsst::afw::image::Wcs`
+            An initial WCS whose CD matrix is used as the final CD matrix and
+            whose CRVAL and CRPIX reference positions are used to setup the
+            grid used when fitting the forward transform from the reverse
+            transform.
+        bbox : :cpp:class:`lsst::afw::geom::Box2I`
+            The region over which the WCS will be valid;
```
TallJimbo: Yes, it's pixels.  Fixed.

python/lsst/meas/astrom/fitSipDistortion.py
TallJimbo: Good catch; fixed.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+            The median on-sky separation between reference objects and
+            sources in "matches", as an lsst.afw.geom.Angle
+        """
+        import lsstDebug
+        display = lsstDebug.Info(__name__).display
+
+        if bbox is None:
+            bbox = lsst.afw.geom.Box2I()
+            for match in matches:
+                bbox.include(lsst.afw.geom.Point2I(match.second.getCentroid()))
```
TallJimbo: Good point; fixed.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+            revFitter.updateModel()
+            intrinsicScatter = revFitter.updateIntrinsicScatter()
+            clippedSigma, nRejected = revFitter.rejectOutliers(self.outlierRejectionCtrl)
+            self.log.debug(
+                "Iteration {0}: intrinsic scatter is {1:4.3f} pixels, "
+                "rejected {2} outliers at {3:3.2f} sigma.".format(
+                    nIter+1, intrinsicScatter, nRejected, clippedSigma
+                )
+            )
+            if display or True:
```
TallJimbo: Indeed it is.  Fixed.

python/lsst/meas/astrom/fitSipDistortion.py
```diff
@@ -0,0 +1,420 @@
...
+        revScaledPoly = revFitter.getTransform()
+        # Convert the generic ScaledPolynomialTransform result to SIP form
+        # with given CRPIX and CD (this is an exact conversion, up to
+        # floating-point round-off error)
+        sipReverse = SipReverseTransform.convert(revScaledPoly, wcs.getPixelOrigin(), cdMatrix)
+
+        # Fit the forward mapping to a grid of points created from the reverse
+        # transform. Because that grid needs to be defined in intermediate
+        # world coordinates, we first grow the bounding box and then apply the
+        # initial WCS's CRPIX and CD matrix.
```
TallJimbo: I'm growing because I'm worried the transform I use to generate the box is only approximate, and I'd rather get a too-big box than a too-small one.  I'll clarify.

python/lsst/meas/astrom/fitSipDistortion.py
TallJimbo: I actually just stole this method (including that doc text) from the old WCS fitter task.  If it's no longer accurate, I'll change it.

python/lsst/meas/astrom/fitSipDistortion.py
TallJimbo: If you think that's the more common pattern, I'm happy to change it.  This is the first time I've actually written interactive display code in a pipeline task.

include/lsst/meas/astrom/PolynomialTransform.h
TallJimbo: Hmm.,  I think I'll just document that instance of this class should be confined to a single thread, as I think that's probably how it ought to be used.  Ultimately I expect it to be a moot point when we switch to using AST for these classes.

However, that's a reminder that the `BinomialMatrix` class I also added on this ticket is unsafe in a somewhat more worrying way, since it uses a singleton.  I'll fix that.

python/lsst/meas/astrom/fitSipDistortion.py
r-owen: It's a slightly delicate issue. Dropping into the debugger is fairly rare, but by no means unheard of. However, the current display code handles frame numbers rather poorly because it does not support any kind of auto-increment. It's easy enough to keep track of a local frame number, but if later code wants to display something in a different frame...

Which reminds me: display is often an integer that states the starting frame number (though if we had auto-increment we'd rarely want that).

python/lsst/meas/astrom/fitSipDistortion.py
r-owen: If it matches current behavior then I agree there seems no strong need to allow keeping them fixed.

src/PolynomialTransform.cc
```diff
@@ -0,0 +1,280 @@
...
+    return compose(
+        scaled.getOutputScalingInverse(),
+        compose(scaled.getPoly(), scaled.getInputScaling())
+    );
+}
+
+PolynomialTransform PolynomialTransform::convert(SipForwardTransform const & other) {
+    PolynomialTransform poly = other.getPoly();
+    poly._xCoeffs(1, 0) += 1;
+    poly._yCoeffs(0, 1) += 1;
```
TallJimbo: Done.  It's to account for the extra terms outside the sum in the SIP definition formula; you can fold those into the sum by adding one to these two coefficients.

src/PolynomialTransform.cc
```diff
@@ -0,0 +1,280 @@
...
+    _xCoeffs(ndarray::allocate(order + 1, order + 1)),
+    _yCoeffs(ndarray::allocate(order + 1, order + 1)),
+    _u(order + 1),
+    _v(order + 1)
+{
+    if (order < 0) {
+        throw LSST_EXCEPT(
+            pex::exceptions::LengthError,
+            "PolynomialTransform order must be >= 0"
+        );
```
TallJimbo: Good catch.  Done (need to move the Eigen initializations too, I think).

src/PolynomialTransform.cc
```diff
@@ -0,0 +1,280 @@
...
+    ndarray::Array<double const,2,2> const & yCoeffs
+) : _xCoeffs(ndarray::copy(xCoeffs)),
+    _yCoeffs(ndarray::copy(yCoeffs)),
+    _u(_xCoeffs.rows()),
+    _v(_xCoeffs.rows())
+{
+    if (xCoeffs.getShape() != yCoeffs.getShape()) {
+        throw LSST_EXCEPT(
+            pex::exceptions::LengthError,
+            "X and Y coefficient matrices must have the same shape."
```
TallJimbo: Done.

src/PolynomialTransform.cc
```diff
@@ -0,0 +1,280 @@
...
+        );
+    }
+    if (_xCoeffs.cols() != _xCoeffs.rows()) {
+        throw LSST_EXCEPT(
+            pex::exceptions::LengthError,
+            "Coefficient matrices must be triangular, not trapezoidal."
```
TallJimbo: For matrices, trapezoidal is to triangular as rectangular is to square.  They're triangular because we don't want nonzero terms with p+q > order, though it's not worth out time to test that.

Offending values printed.

src/PolynomialTransform.cc
```diff
@@ -0,0 +1,280 @@
...
+
+ScaledPolynomialTransform ScaledPolynomialTransform::convert(PolynomialTransform const & poly) {
+    return ScaledPolynomialTransform(poly, afw::geom::AffineTransform(), afw::geom::AffineTransform());
+}
+
+ScaledPolynomialTransform ScaledPolynomialTransform::convert(SipForwardTransform const & sipForward) {
+    ScaledPolynomialTransform result(
+        sipForward.getPoly(),
+        afw::geom::AffineTransform(afw::geom::Point2D() - sipForward.getPixelOrigin()),
+        afw::geom::AffineTransform(sipForward.getCDMatrix())
```
TallJimbo: Fine; done.

src/PolynomialTransform.cc
```diff
@@ -0,0 +1,280 @@
...
+    int const order = t1.getOrder();
+    if (order < 1) {
+        PolynomialTransform t1a(1);
+        t1a._xCoeffs(0, 0) = t1._xCoeffs(0, 0);
+        t1a._yCoeffs(0, 0) = t1._yCoeffs(0, 0);
+        return compose(t1a, t2);
+    }
+    detail::BinomialMatrix binomial(order);
+    // For each of these, (e.g.) a[n] == pow(a, n)
+    auto t2u = detail::computePowers(t2[AT::X], order);
```
TallJimbo: Yes, done.

src/ScaledPolynomialTransformFitter.cc
TallJimbo: Good catch.  I do not want it left true for production code.  I've added an extensive comment.

src/ScaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,465 @@
...
+#include "lsst/afw/table/aggregates.h"
+#include "lsst/afw/table/Match.h"
+#include "lsst/afw/math/LeastSquares.h"
+
+#define LSST_ScaledPolynomialTransformFitter_TEST_IN_PLACE 1
+
+namespace lsst { namespace meas { namespace astrom {
+
+// A singleton struct that manages the schema and keys for polynomial fitting catalogs.
+class ScaledPolynomialTransformFitter::Keys {
```
TallJimbo: It's a similar pattern, though here it has nothing to do with persistence.  I suppose I just have a different instinct when it comes to singletons; if something is logically a singleton (a class with only one or two distinct instances), I tend to make it a singleton in code.  I wonder if @pschella has an opinion on whether this is generally wise or generally foolish?

src/ScaledPolynomialTransformFitter.cc
TallJimbo: Good idea.  Done.

src/ScaledPolynomialTransformFitter.cc
TallJimbo: The relevant documentation for the fields in the schema is already present in the docs for `fromMatches` and `fromGrid` (as well as in the schema itself); since those are the only ways to call the private constructor, I don't think there's any point in documenting the schema there too.

I've added a comment here clarifying that the `valid` field is true if the match is not outlier-rejected, and I'm first checking whether that key is itself valid (i.e. present in the schema).

src/ScaledPolynomialTransformFitter.cc
TallJimbo: Hopefully this is resolved by the above change to disable this block entirely.

src/ScaledPolynomialTransformFitter.cc
TallJimbo: A lambda defined here has access to `ScaledPolynomialTransformFitter`'s private members, which a free function in an anonymous namespace would not.  A private member function would have access to those private members, but then wouldn't be usable as a functor as I could pass to `brent_find_minima`.  So I'd have to have a lambda function anyway to call the private member function, or do a lot of friending.

src/ScaledPolynomialTransformFitter.cc
r-owen: I looked for the docs when I wrote the review and did not find them, so perhaps you could provide a pointer to the help for this function (e.g. "see `fromMatches`")?

src/ScaledPolynomialTransformFitter.cc
TallJimbo: `_keys.valid` is a `Key` for the field that's true if an object is outlier-rejected and false if it isn't.  But if we construct with `fromGrid`, there's no uncertainty and hence no outlier rejection, and that field never gets added to the schema.

(I've added that as a comment).

src/ScaledPolynomialTransformFitter.cc
TallJimbo: Oh, and yes, this should be efficient - `isValid()` is an inline call to a check on whether an integer is positive.

src/ScaledPolynomialTransformFitter.cc
r-owen: I see. And C++ (even C++11) doesn't support defining a function inside a function.

src/SipTransform.cc
TallJimbo: Done.

src/ScaledPolynomialTransformFitter.cc
TallJimbo: I think it does support that, actually, at least in some limited sense, but it's regarded as strongly deprecated in favor of lambdas.

src/ScaledPolynomialTransformFitter.cc
TallJimbo: I think I'm just going to change the name of the field from "valid" to "rejected" and change its sign; I think that will clear up a lot of the confusion.

r-owen: That sounds like a good idea -- rejected or clipped

python/lsst/meas/astrom/fitSipDistortion.py
TallJimbo: I've added separate lsstDebug variables for the frame and whether to pause (I've also seen that as a way to specify the (starting) frame).

tests/testTransforms.py
TallJimbo: Fine.

tests/testTransforms.py
TallJimbo: While it is a long file, I'd prefer to keep these together - I don't think there's anything intrinsically bad about having a few hundred lines of code in one file, and the test utility code is just limited to these tests rather than other stuff in meas_astrom.  And, as we've discussed, I expect all of these classes to eventually be replaced by new AST-like things, and if these tests remain valuable then, I expect they'll be moved between packages and it will all be easier to deal with if it's all in one file.

I will run flake8.

tests/testTransforms.py
TallJimbo: Done.

tests/testTransforms.py
TallJimbo: I've renamed it as you've suggested.

I don't think I'm going to move it somewhere more general right now, though, because I think the domain/range questions are very important for testing transforms more generally, and I don't want to put in the work right now to make this utility function handle more general domains and ranges.

tests/testTransforms.py
TallJimbo: Fixed.  I'll run futurize if I can get it working quickly (haven't tried it before), and I'll certainly run the py3 Jenkins before merging to make sure I haven't missed anything.

tests/testTransforms.py
TallJimbo: Done.

tests/testTransforms.py
TallJimbo: I think you must have misread this - it's already doing exactly what you've asked; the "python transform that applies affine and then poly" is what the first of these lambdas is.

tests/testTransforms.py
TallJimbo: That's actually a lot harder - it requires actually inverting the polynomial part of `rev` (note that `fwd` isn't actually the inverse of `rev` right now).  I do essentially that down in `testFromGrid`.

tests/testTransforms.py
TallJimbo: I'd rather not; it'd only be used in one place, and I think the comments are sufficient to make it clear where in the code we're doing prep work rather than doing the tests.  And I don't want to comb through this function to find which variables I set in the top half I'll need to pass out of the function unless there's a very good reason to split them up.

tests/testTransforms.py
```diff
@@ -0,0 +1,475 @@
...
+            fwd, rev,
+            lsst.afw.coord.IcrsCoord(crval, lsst.afw.geom.degrees)
+        )
+        def t1(p):
+            sky = lsst.afw.coord.IcrsCoord(crval + lsst.afw.geom.Extent2D(p), lsst.afw.geom.degrees)
+            return rev(wcs.skyToIntermediateWorldCoord(sky))
+        def t2(p):
+            sky = lsst.afw.coord.IcrsCoord(crval + lsst.afw.geom.Extent2D(p), lsst.afw.geom.degrees)
+            return wcs.skyToPixel(sky)
+        self.assertTransformsNearlyEqual(t1, t2)
```
r-owen: I had asked if it was also worth copying this test for the reverse direction (using different words) and you replied "That's actually a lot harder - it requires actually inverting the polynomial part of rev (note that fwd isn't actually the inverse of rev right now). I do essentially that down in testFromGrid."

For the record, all I had in mind was literally testing `rev` the same way you test `fwd`, not testing them against each other (I realize they are not inverses of each other, since your documentation was clear on that point). I don't know if it would actually increase test coverage.

TallJimbo: Ah, now I understand.  I think that's a good idea, and I started doing it - but to do it it looks like I'd need a `Wcs` method that doesn't exist: an inverse of `skyTointermediateWorldCoord`.  (And now that I've said that, I can vaguely recall coming to the same conclusion back when I originally wrote this test).

r-owen: That makes sense. May I suggest you add a small comment to that effect?

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
```diff
@@ -0,0 +1,319 @@
...
+ *     rejection and intrinsic scatter estimation.
+ *
+ *   - fromGrid fits an inverse to an existing transform, using the
+ *     to-be-inverted transform to populate a grid of positions.  Because the
+ *     to-be-inverted transform is assumed to be known exactly, fitters
+ *     initialized with this method have no need for outlier rejection or
+ *     scatter estimation.
+ *
+ *  In either case, the fitter creates affine transforms that map the input
+ *  and output data points onto to [-1, 1].  In then fits a polynomial
```
fred3m: Typo: This should be "_**It**_ then fits a polynomial...".

tests/testTransforms.py
TallJimbo: Done.

include/lsst/meas/astrom/ScaledPolynomialTransformFitter.h
TallJimbo: Done.

PR #48, opened 2016-12-14
Only import matplotlib when needed

r-owen:
Some tests and one bit of library code were importing matplotlib
even if plotting was not requested. This could cause needless failures
(not to mention unwanted rebuilds of the font cache).
I standardized on only importing when plotting is specifically requested,
and the library code prints an error message and continues
(whereas the unit tests fail, a safe thing to do because they
do not normally plot anyway).

### DIFF
--- a/python/lsst/meas/astrom/anetBasicAstrometry.py
+++ b/python/lsst/meas/astrom/anetBasicAstrometry.py
@@ -726,9 +726,8 @@ class ANetBasicAstrometryTask(pipeBase.Task):
 
         try:
             import matplotlib.pyplot as plt
-            import numpy
-        except ImportError:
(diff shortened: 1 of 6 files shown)


r-owen:
I have not yet tried the tests with plotting enabled but will do so before merging.

r-owen:
Good suggestion to test the plotting. Setting the back end to `Agg` in `testFitTanSipWcsTask.py` was useful because the plots are saved as pngs so I restored it. I found an error in testLoadAstrometryNetObjects.py's `plotStars` method and fixed that. )Of course all this plotting code is subject to bit rot since it is rarely used.)

r-owen:
I also did what I could to test the plotting changes in the library code. Two of the three files apparently aren't being run by unit tests.

PR #52, opened 2017-01-05
Wrap meas_astrom with pybind11: DM-8458

r-owen:


python/lsst/meas/astrom/SConscript
```diff
@@ -1,4 +1,10 @@
-# -*- python -*-
+## -*- python -*-
```
pschella: remove one `#`

python/lsst/meas/astrom/multiindex.py
```diff
@@ -121,8 +121,7 @@ def read(self):
                 self.log.warn("Unable to get filename for astrometry index %s", fn)
                 continue
             self.log.debug('Reading index from multiindex file "%s"', fn)
-            if multiindex_add_index(self._mi, fn, INDEX_ONLY_LOAD_METADATA):
-                raise RuntimeError('Failed to read index from multiindex filename "%s"' % fn)
+            self._mi.addIndex(fn, False)
```
pschella: does this still raise?

python/lsst/meas/astrom/polynomialTransform.cc
```diff
@@ -0,0 +1,126 @@
...
+                        &ScaledPolynomialTransform::convert,
+                   "other"_a);
+    cls.def_static("convert",
+                   (ScaledPolynomialTransform (*)(SipReverseTransform const &))
+                        &ScaledPolynomialTransform::convert,
+                   "other"_a);
+
+    cls.def("__call__", &ScaledPolynomialTransform::operator(), "in"_a, py::is_operator());
+
+    cls.def("getPoly", &ScaledPolynomialTransform::getPoly, py::return_value_policy::copy);
```
pschella: Are you sure you don't want `py::return_value_policy::reference_internal`?

python/lsst/meas/astrom/polynomialTransform.cc
```diff
@@ -0,0 +1,126 @@
...
+    cls.def("getInputScaling", &ScaledPolynomialTransform::getInputScaling, py::return_value_policy::copy);
```
pschella: ?

python/lsst/meas/astrom/polynomialTransform.cc
```diff
@@ -0,0 +1,126 @@
...
+    cls.def("getOutputScalingInverse", &ScaledPolynomialTransform::getOutputScalingInverse,
+            py::return_value_policy::copy);
```
pschella: ?

python/lsst/meas/astrom/scaledPolynomialTransformFitter.cc
```diff
@@ -0,0 +1,83 @@
...
+    cls.def_static("fromGrid", &ScaledPolynomialTransformFitter::fromGrid);
+
+    cls.def("fit", &ScaledPolynomialTransformFitter::fit, "order"_a=-1);
+    cls.def("updateModel", &ScaledPolynomialTransformFitter::updateModel);
+    cls.def("updateIntrinsicScatter", &ScaledPolynomialTransformFitter::updateIntrinsicScatter);
+    cls.def("getIntrinsicScatter", &ScaledPolynomialTransformFitter::getIntrinsicScatter);
+    cls.def("rejectOutliers", &ScaledPolynomialTransformFitter::rejectOutliers, "ctrl"_a);
+    cls.def("getData", &ScaledPolynomialTransformFitter::getData,
+            py::return_value_policy::reference_internal);
+    cls.def("getTransform", &ScaledPolynomialTransformFitter::getTransform, py::return_value_policy::copy);
```
pschella: Given that you did use `reference_internal` above I assume a copy is actually desired here and in the cases below.

python/lsst/meas/astrom/sip/leastSqFitter1d.cc
```diff
@@ -0,0 +1,76 @@
...
+
+namespace lsst {
+namespace meas {
+namespace astrom {
+namespace sip {
+
+namespace  {
+
+template <typename FittingFunc>
+void declareLeastSqFitter1d(py::module & mod, char const * name) {
```
pschella: Why use `char const *` here but `std::string const &` elsewhere?

python/lsst/meas/astrom/sip/leastSqFitter2d.cc
```diff
@@ -0,0 +1,77 @@
...
+
+namespace lsst {
+namespace meas {
+namespace astrom {
+namespace sip {
+
+namespace  {
+
+template <typename FittingFunc>
+void declareLeastSqFitter2d(py::module & mod, char const * name) {
```
pschella: Same here. I suppose it doesn't matter since we are not composing a suffix and this will save a `std::string` being constructed.

python/lsst/meas/astrom/sip/leastSqFitter2d.cc
```diff
@@ -0,0 +1,77 @@
...
+PYBIND11_PLUGIN(_leastSqFitter2d) {
+    py::module mod("_leastSqFitter2d", "Python wrapper for code in LeastSqFitter2d.h");
+
+    // Need to import numpy for ndarray and eigen conversions
+    if (_import_array() < 0) {
+        PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
+        return nullptr;
+    }
+
+    declareLeastSqFitter2d<afw::math::PolynomialFunction1<double>>(mod, "LeastSqFitter2dPoly");
```
pschella: Actually I think it should be composing, given that `Poly` is clearly a suffix :)

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
+// -*- lsst-C++ -*-
```
pschella: Boilerplate?

python/lsst/meas/astrom/multiindex.py
r-owen: Yes. It calls `MultiIndex.addIndex` (defined in `astrometry_net.cc`), which throws `lsst::pex::exceptions::RuntimeError` if the index cannot be read.

python/lsst/meas/astrom/polynomialTransform.cc
r-owen: The accessors in question return const reference, which I believe we cannot enforce in Python, so I chose to use copy.

python/lsst/meas/astrom/polynomialTransform.cc
r-owen: In this case the returned object is intrinsically const so `reference_internal` does seem to be a better choice. I will change it.

python/lsst/meas/astrom/polynomialTransform.cc
r-owen: Same as below; the returned object is intrinsically const so I'll change it.

python/lsst/meas/astrom/scaledPolynomialTransformFitter.cc
r-owen: In this case I had not wrapped the copy constructor (which I will do) so the object could not be copied. However, this is another case where returning by `reference_internal` is safe.

python/lsst/meas/astrom/sip/leastSqFitter1d.cc
r-owen: In most cases the string will be appended to another, in which case `std::string` is helpful, but not in this case. Nonetheless I will change it for consistency. (I wish `pybind11::class_(...)` accepted `std::string` for the name).

python/lsst/meas/astrom/sip/leastSqFitter2d.cc
r-owen: Exactly. But consistency is probably worth the small loss in efficiency.

python/lsst/meas/astrom/sip/leastSqFitter2d.cc
r-owen: When most of the name is the prefix it seemed cleaner to just provide the full name. I think we tend to over-use prefixes a bit. But I'll change it if you really want.

python/lsst/meas/astrom/astrometry_net.cc
r-owen: I don't understand the question.

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+
+namespace lsst {
+namespace meas {
+namespace astrom {
+
+namespace {
+
+LOG_LOGGER _log = LOG_GET("meas.astrom.astrometry_net");
+
+struct multiindex_deleter {
```
pschella: I think these should be private members of the classes in which they are used.

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+    void operator()(multiindex_t* m) {
+        multiindex_free(m);
+    }
+};
+
+struct solver_deleter {
```
pschella: This one too.

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+    solver_t* s;
+    double timelimit;
+};
+
+static time_t timer_callback(void* baton) {
+    struct timer_baton* tt = static_cast<struct timer_baton*>(baton);
+    solver_t* solver = tt->s;
+    if (solver->timeused > tt->timelimit)
+        solver->quit_now = 1;
+    return 1;
```
pschella: Are you sure this shouldn't be `-1` in case of error?

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+    double getQuadSizeLow() {
+        double qlo,qhi;
+        solver_get_quad_size_range_arcsec(_solver.get(), &qlo, &qhi);
+        return qlo;
+    }
+    double getQuadSizeHigh() {
+        double qlo,qhi;
+        solver_get_quad_size_range_arcsec(_solver.get(), &qlo, &qhi);
+        return qhi;
+    }
```
pschella: This seems to call out for a `std::pair<double, double> getQuadSize()` instead.

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+            }
+
+            solver_add_index(_solver.get(), man.index);
+        }
+    }
+
+    /**
+     * Set parity to flipped (if true) or normal (if false)
+     */
+    void setParity(bool p) {
```
pschella: `setFlippedParity` perhaps more clear?

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+/**
+ * Wrap MultiIndex, a thin shim around multiindex_t
+ */
+void declareMultiIndex(py::module & mod) {
+    py::class_<MultiIndex> cls(mod, "MultiIndex");
+
+    cls.def(py::init<std::string const &>(), "filepath"_a);
+
+    cls.def("__getitem__", &MultiIndex::operator[], "i"_a, py::return_value_policy::reference_internal,
+            py::is_operator());
```
pschella: Note that now it is iterable, but does not satisfy `collections.Iterable`. Perhaps add `__iter__`?

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+    start_an_logging();
+
+    mod.def("healpixDistance", &healpixDistance, "hp"_a, "nside"_a, "coord"_a);
+
+    mod.def("an_log_init", [](int level) {
+        log_init((log_level)level);
+    }, "level"_a);
+
+    mod.def("an_log_get_level", []() {
+        return (int)log_get_level();
```
pschella: `static_cast`

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+    });
+    mod.def("an_log_set_level", [](int level) {
+        log_set_level((log_level) level);
```
pschella: `static_cast`

python/lsst/meas/astrom/astrometry_net.cc
```diff
@@ -0,0 +1,598 @@
...
+                   "Python wrapper for astrometry_net code needed by meas_astrom");
+
+    // code that is run at import time
+    fits_use_error_system();
+    start_an_logging();
+
+    mod.def("healpixDistance", &healpixDistance, "hp"_a, "nside"_a, "coord"_a);
+
+    mod.def("an_log_init", [](int level) {
+        log_init((log_level)level);
```
pschella: `static_cast`

python/lsst/meas/astrom/astrometry_net.cc
r-owen: I am not very knowledgeable about C timer callbacks. However, it appears to me that this function returns 1 whether the time has elapsed or not, and that both conditions are normal. I don't see any code that handles errors.

Note that this is an exact copy of the code that was in `astrometry_net.i`. However, if it needs correcting I'd like to understand better what it needs.

python/lsst/meas/astrom/astrometry_net.cc
r-owen: Agreed. However, I feel that changing the API of the C/C++ should be out of scope, especially for astrometry_net, which we are trying to get rid of. I'm willing to reconsider if you think it important, but my first inclination is to leave it be or file a ticket.

python/lsst/meas/astrom/astrometry_net.cc
r-owen: This could easily be done in Python, but I am not sure how to do it in C++. The existing Python code doesn't use it, so I'm inclined to skip it, since the current code works and we are trying get rid of astrometry_net.

python/lsst/meas/astrom/astrometry_net.cc
r-owen: I thought changing the C++ API was out of scope?

python/lsst/meas/astrom/astrometry_net.cc
r-owen: I changed my mind. It's just used once in Python so I'll rename it to GetQuadSizeRangeArcsec to match the astrometry_net call.

python/lsst/meas/astrom/astrometry_net.cc
r-owen: The current name matches the `astometry_net` call so I'm doubly inclined to leave it alone. However, I will rename the argument `flipped`, which I think will help.

include/lsst/meas/astrom/astrometry_net.h
```diff
@@ -0,0 +1,237 @@
...
+
+    void setMaxStars(int maxStars) {
+        _solver->endobj = maxStars;
+    }
+
+    void setStars(lsst::afw::table::SourceCatalog const & srcs, int x0, int y0);
+
+private:
+
+    struct _Deleter {
```
pschella: Is this our naming convention? Doesn't matter I guess, but just curious.

r-owen: I don't know. It is a private local class and I wanted to make it clear that nobody else can use it.

PR #61, opened 2017-03-21
DM-9812: Clean up outputs from CharacterizeImageTask and CalibrateTask

PaulPrice:
This produces persisted catalogs that are easier to read,
even though they take more space.

### DIFF
--- a/python/lsst/meas/astrom/__init__.py
+++ b/python/lsst/meas/astrom/__init__.py
@@ -44,4 +44,5 @@ from .createMatchMetadata import *
 from .catalogStarSelector import *
 from .directMatch import *
 from .fitSipDistortion import *
+from .denormalizeMatches import *
 from .version import *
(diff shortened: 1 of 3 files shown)


python/lsst/meas/astrom/denormalizedMatches.py
```diff
@@ -0,0 +1,58 @@
+import lsst.afw.table
+
+__all__ = ["matchesToCatalog"]
+
+def matchesToCatalog(matches, matchMeta=None):
+    """Write matches as a Catalog
```
yalsayyad: This could be more descriptive. It doesn't actually write.  How about "Join reference and source catalogs to associative* match list"

* your favorite of associative, cross-reference, junction.

python/lsst/meas/astrom/denormalizedMatches.py
```diff
@@ -0,0 +1,58 @@
+import lsst.afw.table
+
+__all__ = ["matchesToCatalog"]
+
+def matchesToCatalog(matches, matchMeta=None):
```
yalsayyad: I find your module name more descriptive than this function name.   How about "denormalizeMatches"?  Or denormalizeRefMatch?or joinReferenceSource? (since in database lingo, I'd say this is creating a materialized view of the joined reference and source cat dimensions.)

PR #62, opened 2017-03-24
DM-9750: Clean up and stackify matchPessimisticB

morriscb:
DM-9750: Clean up and stackify matchPessimisticB


morriscb:
New version of matcherSourceSelector up that incorporates @r-owen 's comments.

morriscb:
@r-owen, @ctslater "volunteered" to do the majority of the review so don't prioritize this if you could be doing other things. Sorry I initially added so many people to the review.

r-owen:
I would still like a look after you have rebased it.

morriscb:
Rebased to master and squashed commits.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+
+        self.log.warn("Failed after %i patterns." % (pattern_idx + 1))
+        return [], [], None, None
+
+    def _compute_test_vectors(self, source_catalog):
+        """Function to create a list of vectors we can use to test the
+        different rotations against.
+        ------------------------------------------------------
+        """
+        center_vect = np.nanmean(source_catalog, axis=0)
```
r-owen: I suggest a sanity check here. If `source_catalog` is empty you should certainly raise an exception. What if it has only a very few elements? What if all elements of `source_catalog` are `nan`? I suppose you could determine that by testing `center_vect`.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+                               np.degrees(np.arcsin(sin_rot)))
+                return matches, distances, pattern_idx, shift
+
+        self.log.warn("Failed after %i patterns." % (pattern_idx + 1))
+        return [], [], None, None
+
+    def _compute_test_vectors(self, source_catalog):
+        """Function to create a list of vectors we can use to test the
+        different rotations against.
+        ------------------------------------------------------
```
r-owen: Here and for all methods:
- Please document the returned data (in full detail).
- Any method that returns more than one or two values should return a pipe_base Struct or an instance of a custom class. That makes it easier to document what's going on, makes it easier to use the returned data correctly and harder to mess up by getting the order wrong. it also adds some flexibility in that you can return additional fields without breaking anything, and remove fields in a way that reliably breaks code that relied on them (instead of that code mis-assign returned values to the wrong variables and then failing later in "interesting" ways).

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+        self.log = log
+
+        self._build_distances_and_angles()
+
+    def _build_distances_and_angles(self):
+        """ Function to create the data structures we will use to search
+        for our match in.
+        """
+
+        self._id_array = np.empty(
```
r-owen: Please document each data structure somewhere -- perhaps using inline comments where the structure is created. Since they are private variables, that probably makes more sense that adding them to this doc string.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+            sin_rot*ref_cross_matrix +
+            (1. - cos_rot)*np.outer(ref_ctr, ref_ctr))
+
+        shift_rot_matrix = np.dot(rot_matrix, shift_matrix)
+
+        return sin_rot, shift_rot_matrix
+
+    def _intermediate_verify(self, src_pattern, ref_pattern, shift_rot_matrix,
+                             max_dist_rad,):
+        """Preform an intermediate verify step. Rotate the matches references
```
r-owen: Typo: Preform -> Perform.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+                                     ref_pattern.transpose()).transpose()
+        tmp_delta_array = src_pattern - shifted_ref_pattern
+        tmp_dist_array = (tmp_delta_array[:, 0] ** 2 +
+                          tmp_delta_array[:, 1] ** 2 +
+                          tmp_delta_array[:, 2] ** 2)
+        is_good_shift_rot = np.all(tmp_dist_array < max_dist_rad ** 2)
+        return is_good_shift_rot
+
+    def _test_rotation_agreement(self, rot_vect_list, max_dist_rad):
+        """Test this rotation against the pervious N found and return
```
r-owen: Typo: pervious -> previous

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
```diff
@@ -367,9 +375,9 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources, minMa
         @param[in] numUsableSources  number of usable sources (sources with known centroid
             that are not near the edge, but may be saturated)
         @param[in] minMatchedPairs  minimum number of matches
-        @param[in] maxMatchDist  maximum on-sky distance between reference objects and sources
-            (an lsst.afw.geom.Angle); if specified then the smaller of config.maxMatchDistArcSec or
-            maxMatchDist is used; if None then config.maxMatchDistArcSec is used
+        @param[in] toleranceStruct lsst.pipe.base.Struct containing
```
r-owen: I suggest you rename this to not have "Struct" in its name. Perhaps`tolerance`  would suffice? Or `matchTolerance`? This makes a tad more sense if you switch to a custom class, as I hope you will.

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
```diff
@@ -286,6 +287,12 @@ def matchObjectsToSources(self, refCat, sourceCat, wcs, refFluxField, maxMatchDi
             self.log.info("filterStars purged %d reference stars, leaving %d stars" %
                           (preNumObj - numRefObj, numRefObj))
 
+        if toleranceStruct is None:
+            toleranceStruct = pipeBase.Struct(
+                maxMatchDist=None,
+            )
```
r-owen: Here is an excellent example of where a custom class helps. You can replace that with:
```
if tolerance is None:
    tolerance = MatchTolerance()
```
and you get a fully populated instance!

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
```diff
@@ -204,8 +204,8 @@ class MatchOptimisticBTask(pipeBase.Task):
 
     See @ref MatchOptimisticBConfig
 
-    To modify the tests for usable sources and good sources, subclass SourceInfo and
-    set MatchOptimisticBTask.SourceInfoClass to your subclass.
+    To modify the tests for good sources for matching, create a new
+    sourceSelector class in meas_algorithms and use it in the config.
```
morriscb: Took r-owen's first suggestion and replaced the comment text.

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
morriscb: Would this be better to create as its own 1-2 story ticket after this ticket is committed? I could see us creating it and then use it as a slightly bigger restructure that stores not just the tolerances, but the matches themselves and methods for computing statistics on the matches.

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
morriscb: It's a habit of mine to name the ending of a variable it's type. Certainly if we decide to go with the MatchObject or MatchTolerance class we can do this but as is I like the type to be in the name.

python/lsst/meas/astrom/ref_match.py
```diff
@@ -27,6 +27,7 @@
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 from .matchOptimisticB import MatchOptimisticBTask
+from .matchPessimisticB import MatchPessimisticBTask
```
morriscb: Right, this was left over from when I was testing and before I discovered the retarget command in config files. I have removed the MatchPessimisticBTask import.

tests/testMatchPessimisticB.py
```diff
@@ -0,0 +1,216 @@
...
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+import math
+import os
+import unittest
+
+import lsst.daf.base as dafBase
+import lsst.pipe.base as pipeBase
```
morriscb: Removed.

tests/testMatchPessimisticB.py
```diff
@@ -0,0 +1,216 @@
...
+        metadata.set("CRVAL1", 36.930640)
+        metadata.set("CRVAL2", -4.939560)
+        metadata.set("CRPIX1", 792.4)
+        metadata.set("CRPIX2", 560.7)
+        metadata.set("CD1_1", -5.17e-05)
+        metadata.set("CD1_2", 0.0)
+        metadata.set("CD2_2", 5.17e-05)
+        metadata.set("CD2_1", 0.0)
+        self.wcs = afwImage.makeWcs(metadata)
+        self.distortedWcs = self.wcs
```
morriscb: self.distoredWcs is a copy at this point but if you look at the later tests, which are the same as testMatchOptimisticB it does end up being set to a distorted wcs in testLargeDistortion and for the individual distort tests (e.g. testLinearXDistort). PS I didn't add this test, it must be showing up incorrectly in the commit history.

This sounds like another ticket to make a better test for both MatchOptimistic and PessimsiticB. I'm happy to collaborate and work on it. I'm not going to be as familiar with all of the cases we want to test for optical distortion.

tests/testPPMb3D.py
```diff
@@ -0,0 +1,228 @@
+
```
morriscb: Renamed to the full unabbreviated name.

tests/testPPMb3D.py
```diff
@@ -0,0 +1,228 @@
...
+    """Unittest suite for the python implimentation of
+    Optimistic Pattern Matcher B.
+    """
+
+    def setUp(self):
+        np.random.seed(12345)
+
+        n_points = 1000
+        self.reference_catalog = np.empty((n_points, 4))
```
morriscb: Renamed to reference_obj_array and source_obj_array respectively.

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
r-owen: We discourage that naming convention in LSST code. I stretch it a bit sometimes to include "Ptr" for pointers (in C++ code) but even that has raised complaints so I try to use it judiciously. I strongly recommend avoiding "Struct" in this variable name.

tests/testPPMb3D.py
```diff
@@ -0,0 +1,228 @@
...
+        self.reference_catalog[:, 0] = sin_theta_array*np.cos(phi_array)
+        self.reference_catalog[:, 1] = sin_theta_array*np.sin(phi_array)
+        self.reference_catalog[:, 2] = cos_theta_array
+        self.reference_catalog[:, 3] = (
+            np.random.power(1.2, size=n_points)*4 + 20)
+        self.source_catalog = copy(self.reference_catalog)
+        self.log = Log()
+
+    def testInit(self):
+        self.pyPPMb = PessimisticPatternMatcherB(
```
morriscb: Removed init test.

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
r-owen: I suggest adding the class now and expanding it on a separate ticket.

tests/testPPMb3D.py
```diff
@@ -0,0 +1,228 @@
...
+            reference_catalog=self.reference_catalog[:, :3],
+            log=self.log)
+
+    def testConstructAndMatchPattern(self):
```
morriscb: Changed test name to testConstructPattern.

tests/testPPMb3D.py
```diff
@@ -0,0 +1,228 @@
...
+            np.cos(np.radians(60. / 3600.)), np.cos(np.radians(1.0)) ** 2,
+            np.radians(5./3600.))
+        self.assertEqual(pattern_list[0], 2)
+        self.assertEqual(pattern_list[1], 4)
+        self.assertEqual(pattern_list[2], 8)
+        self.assertEqual(pattern_list[3], 16)
+        self.assertEqual(pattern_list[4], 32)
+        self.assertEqual(pattern_list[5], 64)
+
+    def testMatchPerfect(self):
```
morriscb: Doc string added to this and all tests.

tests/testPPMb3D.py
```diff
@@ -0,0 +1,228 @@
...
+            log=self.log)
+
+        matches, distances, pattern_idx, shift = self.pyPPMb.match(
+            source_catalog=self.source_catalog[:500], n_check=9, n_match=6,
+            n_agree=2, max_n_patterns=100, max_shift=60., max_rotation=1.0,
+            max_dist=5., min_matches=30, pattern_skip_array=None)
+        self.assertEqual(len(matches), len(self.reference_catalog[:500]))
+        self.assertTrue(np.all(distances < 0.01/3600.0*__deg_to_rad__))
+
+    def testShift(self):
```
morriscb: This rotation of the unit sphere will appear as a rotation on our "focal plane". Added text to the doc string explaining this.

r-owen: Thanks. (But if a rotation on one appears as a rotation on the other then the name `testShift` is still puzzling. Perhaps you meant to say that a rotation on the one appears as a shift on the other).

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+
+
+class PessimisticPatternMatcherB(object):
+    """Class implimenting a pessimistic version of Optimsitic Pattern Matcher B
+    from Tabur 2007. The class loads and stores the reference object in a
+    convienent data structure for matching any set of source obejcts that are
+    assumed to contain each other. The pessimistic nature of the algorithm
+    comes from requiring that it discovers at least two patterns that agree on
+    the correct shift and rotation for matching before exiting. The original
+    behavior of OPMb can be recovered simply.
```
morriscb: Changed all "function" descriptions as described. Added a short description of what a pattern to be matched looks like and referred the interested party to the current DMTN that I am currently writing.

tests/testPPMb3D.py
morriscb: I misspoke. I should have edited my reply. What I mean to say was "This rotation of the unit sphere will appear as a shift on our "focal plane". Added text to the doc string explaining this."

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Added comments throughout the construction of the matching data structures. I also restructured the code based on comments from @ctslater.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+                self._dist_array, tmp_sort_dist_array)
+
+        self._id_array = self._id_array[self._sorted_args]
+        self._delta_array = self._delta_array[self._sorted_args]
+
+    def match_optimistic(self, source_catalog, n_check, n_match,
+                         max_n_patterns, max_shift, max_rotation, max_dist,
+                         min_matches, pattern_skip_array=None):
+        """Wrapper for the match method that sets the n_agree variable to 1,
+        recoverying the behavior of an optimistic pattern matcher.
```
morriscb: Removed match_optimistic from the API and added a comment to n_agree.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+            max_shift: float maximum allowed shift to match patterns in
+                arcseconds.
+            max_rotation: float maximum allowed rotation between patterns in
+                degrees.
+            max_dist: float value specifying the maximum distance in
+                arcseconds allowed between candidate spokes in the source and
+                reference objects. Also sets that maximum distance in the
+                intermediate verify, pattern shift/rotation agreement, and
+                final verify steps.
+            pattern_skip_array: int array of patterns we would like to skip.
```
morriscb: I added a bit of text to the comments describing this, but you are correct in each of your points. Setting the pattern skip is based on pervious iterations of the matcher where it is suspected that the previous pattern matched was incorrect. This usually happens because, on a second matchAndFit iteration, the code has to soften it's match criteria.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+                This could be due to the pattern being matched on a
+                pervious iteration that we now consider invalid.
+
+        Returns:
+            tuple (2D int array of matched pairs,
```
morriscb: I added a bit of text explaining the output. Once we finish will all of the more minor comments I'll start on the major ones if we need them like turning this output into a struct or creating a class for the tolerance.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+        # To test if the shifts and rotations we find agree we first create
+        # two test points situtated at the top and bottom of where the z axis
+        # on the sphere bicects the source catalog.
+        test_vect_list = self._compute_test_vectors(source_catalog[:, :3])
+
+        # We now create an empty list of our resultant rotator vectors to
+        # compare the different roations we find.
+        rot_vect_list = []
+        # Dummy variable for storing and returning the index of the pattern
+        # we succeeded in matching.
```
morriscb: I have removed the later comment as it was redundant with that above.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,722 @@
...
+                                          n_source - n_match))):
+            # More Dummy storage variables for the matches we return and
+            matches = []
+            distances = []
+            ref_candidates = []
+            src_candidates = []
+            shift_rot_matrix = None
+            # If this pattern is one that we matched on the past but we
+            # now want to skip it, we can do so here.
+            if not (pattern_skip_array is None) and \
```
morriscb: Fixed.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Fixed.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Fixed.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Changed multiple outputs to pipeBase structs. The output of _compute_test_vectors remains unchanged however as I do want a list of arrays to use in later tests. I have added a few comments to clarify this.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Added a warning if any of the source catalog values are non-finite. Also added an early exit and warning if the input source array is empty.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+
+        # Our first test. We search the reference dataset for pairs
+        # that have the same length as our first source pairs to with
+        # plus/minus the max_dist tolerance.
+        ref_dist_index_array = self._find_candidate_reference_pairs(
+            src_dist_array[0], self._dist_array, max_dist_rad)
+
+        # Start our loop over the candidate reference objects.
+        for ref_dist_idx in ref_dist_index_array:
+            # We hav two candidates for which reference object corresponds
```
r-owen: "hav" -> "have"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            # with the source at the center of our pattern. As such we loop
+            # over and test both possiblities.
+            tmp_ref_pair_list = self._id_array[ref_dist_idx]
+            for pair_idx, ref_id in enumerate(tmp_ref_pair_list):
+                matched_src_list = [0, 1]
+                matched_ref_list = []
+                # Test the angle between our candidate ref center and the
+                # source center of our pattern. This angular distance also
+                # definse the shift we will later use.
```
r-owen: "definse" -> "defines"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+                if len(ref_spoke_list) < n_match - 2 or \
+                   len(src_spoke_list) < n_match - 2:
+                    continue
+
+                # If we have the right number of matched ids we store these.
+                matched_ref_list.extend(ref_spoke_list)
+                matched_src_list.extend(src_spoke_list)
+
+                # We can now create our full rotation matrix for both the
+                # shift and rotation. Reminder shift, alings the pattern
```
r-owen: "alings" -> "aligns"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+                # If we fail to create the rotation matrix, continue to the
+                # next objects.
+                if shift_rot_struct is None:
+                    continue
+
+                # Get the data from the return struct.
+                sin_rot = shift_rot_struct.sin_rot
+                shift_rot_matrix = shift_rot_struct.shift_rot_matrix
+
+                # Now that we have enough canidates we test to see if it passes
```
r-owen: "canidates" -> "candidates"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+                # intermediate verify. This shifts and rotates the soruce
```
r-owen: "soruce" -> "source"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+                        src_candidates=matched_src_list,
+                        shift_rot_matrix=shift_rot_matrix,
+                        cos_shift=cos_shift,
+                        sin_rot=sin_rot)
+        return None
+
+    def _find_candidate_reference_pairs(self, src_dist, ref_dist_array,
+                                        max_dist_rad):
+        r"""Wrap numpy.searchsorted to find the range of reference spokes within
+        a spoke distance tolernace of our source spoke. Returns an array sorted
```
r-owen: Brief description followed by blank line followed by details. (Here and in all doc strings)

"tolernace" -> "tolerance"

Returns -> Return (active voice) (here and in all doc strings)

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            sorted array of distances in radians.
+        max_dist_rad : float
+            maximum plus/minus search to find in the reference array in
+            radians.
+
+        Returns
+        -------
+        int array
+            indices lookup into the input ref_dist_array sorted by the
+            difference in value to the src_dist from absploute value
```
r-owen: "absploute" -> "absolute"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        if start_idx == end_idx:
+            return []
+
+        # Make sure the endpoints of the input array are respected.
+        if start_idx < 0:
+            start_idx = 0
+        if end_idx > ref_dist_array.shape[0]:
+            end_idx = ref_dist_array.shape[0]
+
+        # Now we sort the indicies from smallest absolute delta dist difference
```
r-owen: "indicies" -> "indices"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+                       cos_shift, max_cos_rot_sq):
+        r""" Test if the rotation implied between the source
+        pattern and reference pattern is within tolerance. To test this
+        we need to create the first part of our spherical rotation matrix
+        which we also return for use later.
+
+        Parameters
+        ----------
+        src_center : float array
+            3 vector defining the center of our current source pin wheel
```
r-owen: "pin wheel" -> "pinwheel"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        src_delta : float array
+            3 vector delta between the source pattern center and the end of
+            the pinwheel spoke.
+        ref_delta : float array
+            3 vector delta of the candidate matched reference pair
+        cos_shift : float
+            Cosine of the angle between the source and reference candidate
+            centers.
+        max_cos_rot_sq : float
+            Maximum amount of rotation allowed bweteen the source pair and our
```
r-owen: "btweteen"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            candidate reference pair after shifting the centers on top of each
+            other. The function will return None if the rotation implied is
+            greater than max_cos_rot_sq.
+
+        Returns
+        -------
+        None or lsst.pipe.base.Struct
+            If the rotation implied by the source and reference candidate pairs
+            is greater than max_cos_rot_sq we return None. If it is less we
+            return a pipe.base.Struct containing the following data.
```
r-owen: Another candidate for always returning a Struct

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            return None
+        return pipeBase.Struct(
+            cos_rot_sq=cos_rot_sq,
+            shift_matrix=shift_matrix,)
+
+    def _create_pattern_spokes(self, src_ctr, src_delta_array, src_dist_array,
+                               ref_ctr, ref_ctr_id, ref_delta, ref_dist,
+                               ref_delta_array, ref_dist_array,
+                               ref_id_array, max_dist_rad, n_match):
+        """ Create the indiviual spokes that make up the pattern now that the
```
r-owen: "indiviual"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            3 vector of the source pinwheel center
+        src_delta_array : float array
+            Array of 3 vector deltas between the source center and the pairs
+            that make up the remaining spokes of the pinwheel
+        src_dist_array : float array
+            Array of the distances of each src_delta in the pinwheel
+        ref_ctr : float array
+            3 vector of the candidate refenerce center
+        ref_ctr_id : int
+            id lookup of the ref_ctr into the master reference array
```
r-owen: I think this would be clearer as "ID of the ref_ctr in the master reference array" (if that's what it is).

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        ref_dist : float
+            Radian distance of the first candidate reference pair
+        ref_delta_array : float array
+            Array of 3 vector deltas that are have the current candidate
+            reference center as part of the pair
+        ref_dist_array : float array
+            Array of vector distances for each of the reference pairs
+        ref_id_array : int array
+            Array of pair id lookups into the master reference array. All
+            pairs contain the reference center id.
```
r-owen: Might you clarify this? I really don't get it. I think it's a pair of IDs of master reference array objects, but if so, it's less clear why they are paired.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        max_dist_rad : float
+            Maximum search distance
+        n_match : int
+            Number of source deltas that must be matched into the reference
+            deltas in order to consider this a successful pattern match.
+
+        Returns
+        -------
+        None or lsst.pipe.base.Struct
+            None if no pattern match is found. Return a pipe.base.Struct
```
r-owen: Another case for always Struct?

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            src_sin_tol = (max_dist_rad /
+                           (src_dist_array[src_idx] + max_dist_rad))
+            # Test if the small angle approximation will still hold. This is
+            # defined as when sin(theta) ~= theta to within 0.1% of each other.
+            # This also implicitly sets a minimum spoke length that we can use.
+            if src_sin_tol > 0.0447:
+                n_fail += 1
+                continue
+
+            # Find the reference pairs that include our candiates pattern
```
r-owen: "candiate"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+                src_ctr, src_delta_array[src_idx], src_dist_array[src_idx],
+                src_delta_array[0], src_dist_array[0], ref_ctr, ref_ctr_id,
+                ref_delta, ref_dist, ref_dist_idx_array, ref_delta_array,
+                ref_dist_array,
+                ref_id_array, src_sin_tol)
+            if ref_id is None:
+                n_fail += 1
+                continue
+
+            # Append the successful indicies to our list. The src_idx needs
```
r-owen: "indicies" -> "indices"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+    # the source dot and cross products of the opening angle in the
+    # previous method.
+    def _test_spoke(self, src_ctr, src_delta, src_dist, src_ctr_delta,
+                    src_ctr_dist, ref_ctr, ref_ctr_id, ref_delta, ref_dist,
+                    ref_dist_idx_array, ref_delta_array, ref_dist_array,
+                    ref_id_array, src_sin_tol):
+        """Test the opening angle between the first spoke of our pattern
+        for the soruce object against the reference object. This method makes
+        heavy use of the small angle approximation assumption to perform the
+        comparison.
```
r-owen: Separate brief description from rest of description

"soruce"

Consider eliding "assumption" to give: "heavy use of the small angle approximation to perform the comparison"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        src_dist : float array
+            Distance of the current spoke we are testing
+        src_ctr_delta : float array
+            3 vector delta between the center of the pattern and the first
+            spoke of the pattern. Used to test compute the opening angle
+            between the current spoke and the first spoke.
+        src_ctr_dist : float
+            Distance between the pairs that make up src_ctr_delta
+        ref_ctr : float array
+            3 vector of the candidate refenerce center
```
r-owen: "refenerce"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        ref_ctr_id : int
+            id lookup of the ref_ctr into the master reference array
+        ref_delta : float array
+            3 vector of the first candidate pair of the pinwheel. That is
+            the candidate pair that was matched in the
+            main _construct_pattern_and_shift_rot_matrix loop
+        ref_dist : float
+            Radian distance of the first candidate reference pair
+        ref_dist_idx_array : int array
+            Indicies sorted by the delta distance between the source
```
r-owen: "Indicies" -> "Indices" (this is probably be worth a global search)

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            pairs contain the reference center id.
+        src_sin_tol : float
+            Sine of tolerance allowed between source and reference spoke
+            opening angles.
+
+        Returns
+        -------
+        None or int
+            If we can not find a candidate spoke we return None else we
+            return an int id into the master reference array.
```
r-owen: Just for the record, I think this is a perfect place to return None

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        Returns
+        -------
+        lsst.pipe.base.Struct
+            Struct object containing the following data
+            sin_rot : float sine of the amount of rotation between the
+                source and reference pattern. We use sine here as it is
+                signed and tells us the chirality of the rotation.
+            shift_rot_matrix : float array representing the 3x3 rotation
+                matrix that takes the source patern and shifts and rotates
+                it to align with the reference pattern.
```
r-owen: Very nice descriptions, but I think each description needs to start on the line *after* the name and type

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            maximum distance to consider an object a match after rotating the
+            the source and references into the other's frame.
+
+        Returns
+        -------
+        lsst.pipe.base.Struct
+            A Struct object containing the following data
+            matches : a (N, 2) array of integer ids into the source and
+                reference arrays. Matches are only returned for those that
+                staisfy the distance and handshake criteria.
```
r-owen: "staisfy"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        final_mask = np.logical_and(handshake_mask,
+                                    tmp_src_dist < max_dist_rad)
+        return pipeBase.Struct(
+            matches=ref_matches[final_mask],
+            distances=tmp_src_dist[final_mask],)
+
+    def _handshake_match(self, matches_ref, matches_src):
+    	"""Return only those matches where both the source
+        and reference objects agree they they are each others'
+        nearist neighbor.
```
r-owen: "nearist"

tests/test_MatchPessimisticB.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python
```
r-owen: No `#!` line on unit tests. Sorry, but it breaks on Macs and so has been forbidden by our masters

tests/test_MatchPessimisticB.py
```diff
@@ -0,0 +1,215 @@
...
+        if doPlot:
+            import matplotlib.pyplot as plt
+            undistorted = [self.wcs.skyToPixel(self.distortedWcs.pixelToSky(ss.getCentroid())) for
+                           ss in distortedCat]
+            refs = [self.wcs.skyToPixel(ss.getCoord()) for ss in refCat]
+
+            def plot(catalog, symbol):
+                plt.plot([ss.getX() for ss in catalog], [ss.getY() for ss in catalog], symbol)
+
+            # plot(sourceCat, 'k+') # Original positions: black +
```
r-owen: Folks tend to get shirty about commented-out code. If you really don't find this useful can you delete it? Otherwise please explain why it's commented out and when it can be removed or re-enabled.

tests/test_MatchPessimisticB.py
```diff
@@ -0,0 +1,215 @@
...
+    def loadSourceCatalog(self, filename):
+        """Load a list of xy points from a file, set coord, and return a SourceSet of points
+        """
+        sourceCat = afwTable.SourceCatalog.readFits(filename)
+        aliasMap = sourceCat.schema.getAliasMap()
+        aliasMap.set("slot_ApFlux", "base_PsfFlux")
+        fluxKey = sourceCat.schema["slot_ApFlux_flux"].asKey()
+        fluxSigmaKey = sourceCat.schema["slot_ApFlux_fluxSigma"].asKey()
+
+        # print("schema=", sourceCat.schema)
```
r-owen: Please remove this. Folks can add it back temporarily if they need it

tests/test_PessimisticPatternMatcherB3D.py
```diff
@@ -0,0 +1,273 @@
+
```
r-owen: Please add the usual boilerplate

tests/test_PessimisticPatternMatcherB3D.py
```diff
@@ -0,0 +1,273 @@
...
+        # Our initial source catalog is a straight copy of the reference
+        # array at first. In some of the tests we will add rotations and
+        # shifts to the data in order to test the input and outputs of our
+        # matcher.
+        self.source_obj_array = copy(self.reference_obj_array)
+        self.log = Log()
+
+    def testConstructPattern(self):
+        """ Test that a specified pattern can be found in the refernce
```
r-owen: "refernce"

tests/test_PessimisticPatternMatcherB3D.py
```diff
@@ -0,0 +1,273 @@
...
+            log=self.log)
+
+        match_struct = self.pyPPMb.match(
+            source_catalog=self.source_obj_array, n_check=9, n_match=6,
+            n_agree=2, max_n_patterns=100, max_shift=60., max_rotation=5.0,
+            max_dist=5., min_matches=30, pattern_skip_array=None)
+        self.assertEqual(len(match_struct.matches),
+                         len(self.reference_obj_array))
+        self.assertTrue(
+            np.all(match_struct.distances < 0.01/3600.0 * __deg_to_rad__))
```
r-owen: Consider using numpy.testing.assert_all; also please use named constants for clarity (here and elsewhere). For example:

```
from numpy.testing import assert_all
...
maxAllowedDistRad = 0.01/3600.0 * __deg_to_rad__
assert_all(match_struct.distances < maxAllowedDistRad)
```

tests/test_PessimisticPatternMatcherB3D.py
```diff
@@ -0,0 +1,273 @@
...
+            max_dist=5., min_matches=30, pattern_skip_array=None)
+        self.assertEqual(len(match_struct.matches),
+                         len(self.reference_obj_array[:500]))
+        self.assertTrue(
+            np.all(match_struct.distances < 0.01/3600.0 * __deg_to_rad__))
+
+    def testShift(self):
+        """ Test the matcher when a shift is applied to the data. We say
+        shift here as while we are rotating the unit-sphere in 3D on our
+        'focal plane' this will appear as a shift.
```
r-owen: Very nice explanation (though again...please have a brief description followed by a blank line, followed by details)

tests/test_PessimisticPatternMatcherB3D.py
```diff
@@ -0,0 +1,273 @@
...
+            n_agree=2, max_n_patterns=100, max_shift=60, max_rotation=5.0,
+            max_dist=5., min_matches=30, pattern_skip_array=None)
+
+        self.assertEqual(len(match_struct.matches),
+                         len(self.reference_obj_array))
+        self.assertTrue(
+            np.all(match_struct.distances < 0.1/3600.0 * __deg_to_rad__))
+
+    def testRotation(self):
+        """ Test the matcher for when a roation is applied to the data.
```
r-owen: "roation"

tests/test_PessimisticPatternMatcherB3D.py
```diff
@@ -0,0 +1,273 @@
...
+        self.pyPPMb = PessimisticPatternMatcherB(
+            reference_catalog=self.reference_obj_array[:, :3],
+            log=self.log)
+        theta = np.radians(45.0 / 3600.)
+        cos_theta = np.cos(theta)
+        sin_theta = np.sin(theta)
+        theta_rotation = np.array(
+            [[cos_theta, -sin_theta, 0.],
+             [sin_theta,  cos_theta, 0.],
+             [       0.,         0., 1.]])
```
r-owen: You use this more than once. How about adding a method to compute it? Similarly for some other rotation matrices.

python/lsst/meas/astrom/matchOptimisticB/matchOptimisticBContinued.py
```diff
@@ -367,9 +309,9 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources, minMa
         @param[in] numUsableSources  number of usable sources (sources with known centroid
             that are not near the edge, but may be saturated)
         @param[in] minMatchedPairs  minimum number of matches
-        @param[in] maxMatchDist  maximum on-sky distance between reference objects and sources
-            (an lsst.afw.geom.Angle); if specified then the smaller of config.maxMatchDistArcSec or
-            maxMatchDist is used; if None then config.maxMatchDistArcSec is used
+        @param[in] toleranceStruct lsst.pipe.base.Struct containing
+            variables specifying matcher tolerances and state from possible
+            previous runs.
```
morriscb: Removed as it was not used in the new MatchTolerance implementation.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+        max=6.0,
+    )
+    numPointsForShape = pexConfig.Field(
+        doc="Number of points to define a shape for matching.",
+        dtype=int,
+        default=6,
+    )
+    numPointsForShapeAttempt = pexConfig.Field(
+        doc="Number of points to try for creating a shape. This value should "
+            "be greater than numPointsForShape.",
```
morriscb: Implemented a validate method for this config comparison.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+            "saturated objects, edge objects, and interpolated objects.",
+        default="matcher"
+    )
+
+    def setDefaults(self):
+        sourceSelector = self.sourceSelector["matcher"]
+        sourceSelector.setDefaults()
+
+    # The paremeters below are not used in this matcher and are kept around
+    # for compatability.
```
morriscb: Here's the problem with removing these through. Right now the processCcd.py config for obs_subaru/hsc sets these parameters by hand. Removing them here would cause breaks for the pessimistic task that would be difficult to track down. Once we do the RFC to replace the current matcher I'm happy to remove them and get rid of the defaults in obs_subaru but for now to keep things as they are I'm okay with having them stick around considering we will have an upcoming ticket/epic to change to the Pessimistic matcher.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+    ConfigClass = MatchPessimisticBConfig
+    _DefaultName = "matchObjectsToSources"
+
+    def __init__(self, **kwargs):
+        pipeBase.Task.__init__(self, **kwargs)
+        self.makeSubtask("sourceSelector")
+
+    def filterStars(self, refCat):
+        """Extra filtering pass; subclass if desired
+        """
```
morriscb: I have no idea. This is left over from the API of MatchOptimsiticB. I'm removing it from MatchPessimsitic as of now.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+            state variables from the match.
+        """
+        import lsstDebug
+        debug = lsstDebug.Info(__name__)
+
+        preNumObj = len(refCat)
+        refCat = self.filterStars(refCat)
+        numRefObj = len(refCat)
+
+        if self.log:
```
morriscb: Left over from MatchOptimisticB. The full line is now removed as I do not use a star filter.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+
+        # avoid accidentally using sourceCat; use goodSourceCat from now on
+        del sourceCat
+
+        minMatchedPairs = min(self.config.minMatchedPairs,
+                              int(self.config.minFracMatchedPairs *
+                                  min([len(refCat), len(goodSourceCat)])))
+
+        # match usable (possibly saturated) sources and then purge saturated
+        # sources from the match list
```
morriscb: The comment has been removed. As you say, I no longer use saturated objects in the matcher.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+
+        # Set configurable defaults when we encounter None type or set
+        # state based on pervious run of matchandfit.
+        if match_tolerance.maxShift is None:
+            maxShiftArcseconds = (self.config.maxOffsetPix *
+                                  wcs.pixelScale().asArcseconds())
+        else:
+            maxShiftArcseconds = np.max(
+                (match_tolerance.maxShift.asArcseconds(),
+                 wcs.pixelScale().asArcseconds()))
```
morriscb: A comment has been added to clarify that, yes, this is a test to keep the max shift allowed from falling below 1 pixel in size.

python/lsst/meas/astrom/matchPessimisticB.py
r-owen: I figured it was something like that. However, I don't feel that justifies leaving them around. We have to deal with this sort of thing all the time and we've gotten used to it.

r-owen: I suggest asking about it on the `dm` Slack room? If you can convince those folks then I certainly won't stand in the way (beyond suggesting you file a ticket for future removal).

PaulPrice: I think you should remove them.  If you activate this `Task` in the correct way, you won't hit any problems.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+            tmp_ra = coord.getLongitude().asRadians()
+            tmp_dec = coord.getLatitude().asRadians()
+            theta = np.pi/2 - tmp_dec
+            phi = tmp_ra
+            src_array[src_idx, 0] = np.sin(theta)*np.cos(phi)
+            src_array[src_idx, 1] = np.sin(theta)*np.sin(phi)
+            src_array[src_idx, 2] = np.cos(theta)
+            src_array[src_idx, 3] = -2.5*np.log10(
+                np.where(srcObj.getPsfFlux() > 0,
+                         srcObj.getPsfFlux(), 10**-32))
```
morriscb: Abstracted the flux and xyz conversion into a function. Accessors to flux and RA, DEC are not consistent between ReferenceCatalog and SourceCatalog so those calls remain.

python/lsst/meas/astrom/matchPessimisticB.py
morriscb: Okay, they have been removed.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -0,0 +1,594 @@
...
+                    max_dist=maxMatchDistArcSec * 2. ** try_idx,
+                    min_matches=minMatchedPairs,
+                    pattern_skip_array=np.array(
+                        match_tolerance.failedPatternList)
+                )
+            if matcher_struct is None and \
+               match_tolerance.lastPattern is None and \
+               try_idx == 0:
+                # If we found a pattern on a previous run and can't
+                # find an optimistic match with the harhest tolerances
```
morriscb: Corrected the typo and added the comma as suggested. You did read the comment correctly.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+        Returns
+        -------
+        output_struct : pipe.base.struct
+            A pipebase struct containing the following outputs
+                maches : a (N, 2) int array of matched pairs
+                distances : a float array of pair distances
+                pattern_idx : int index of matched pattern
+                shift : a float value for the shift between the source and
+                    reference objects.
+        Returns None type if no match is found.
```
morriscb: I don't really need to return partial results as the matcher is pretty all or nothing. Using None as a return on failure actually simplified the tests in the code a bit as I can just test of the output struct exists, that is the test succeeded. I did change the text as you suggested through.

morriscb: Actually, never mind. I have thought of a way this is useful. I will make the change to only return a struct.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -0,0 +1,1111 @@
...
+            to have their centers on top of each other.
+        max_dist_rad : float
+            Maximum delta distance allowed between the source and reference
+            pair distances to consider the reference pair a candidate for
+            the source pair. Also sets the tolerance between the opening
+            angles of the spokes when compared to the reference.
+
+        Returns
+        -------
+        None or lsst.pipe.base.Struct
```
morriscb: There are no cases where I really want to compute partial data. In the case of the matcher, return a partial result could actually be worse than returning nothing. I understand how this could work in other code, however I would rather have a hard failure that is easy to track than a soft one that can act like a success when in fact the results are extremely wrong. I've started using the structs as requested but on failures I set them to empty lists or None type as in most places I do not want a partial result.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: I suppose I could also check to see if the candidate id is less than that of the id of the center, as that would tell me the direction of the delta vector. This reduces the size of the pair_id_array by half so seems like an upgrade. Change made and doc altered.

tests/test_PessimisticPatternMatcherB3D.py
morriscb: Created a class function in PessimisticPatternMatcher that computes an arbitrary 3D rotation matrix.

tests/test_PessimisticPatternMatcherB3D.py
morriscb: In the future Russell, could you please be more clear as to what you mean. I haven't been working in LSST long enough to know what the usual boilerplate means.

tests/test_MatchPessimisticB.py
morriscb: In that case someone better start cleaning up meas_astrom fast because the majority of the unittests have #!. This there currently because the test was copied from MatchOptimisticB. It is now removed.

PR #67, opened 2017-04-25
Add calib_astrometryUsed field to record sources used

pgee2000:
Also added a doWrite config field to match photoCal and measurePsf

### DIFF
--- a/python/lsst/meas/astrom/astrometry.py
+++ b/python/lsst/meas/astrom/astrometry.py
@@ -22,6 +22,8 @@
 from __future__ import absolute_import, division, print_function
 from builtins import range
 
+import random
+
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 from .ref_match import RefMatchTask, RefMatchConfig
@@ -54,6 +56,11 @@ class AstrometryConfig(RefMatchConfig):
         dtype=float,
         default=0.001,
         min=0,
+    )
+    doWriteOutput = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Write used flag to the schema",
     )
 
 # The following block adds links to this task from the Task Documentation page.
@@ -147,6 +154,13 @@ class AstrometryTask(RefMatchTask):
         @param[in] kwargs  additional keyword arguments for pipe_base Task.\_\_init\_\_
         """
         RefMatchTask.__init__(self, refObjLoader, schema=schema, **kwargs)
+
+        if self.config.doWriteOutput and not schema is None:
+            self.usedKey = schema.addField("calib_astrometryUsed", type="Flag",
+                                               doc="set if source was used in astrometric calibration")
+        else:
+            self.usedKey = None
+
         self.makeSubtask("wcsFitter")
 
     @pipeBase.timeMethod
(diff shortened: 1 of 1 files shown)


PaulPrice:
But that `Task` doesn't write anything, so the config field is a bit misleading.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -55,6 +57,11 @@ class AstrometryConfig(RefMatchConfig):
         default=0.001,
         min=0,
     )
+    doWriteOutput = pexConfig.Field(
```
parejkoj: I feel like a better name for this config parameter would be `doOutputUsedFlag`.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -147,6 +154,13 @@ def __init__(self, refObjLoader, schema=None, **kwargs):
         @param[in] kwargs  additional keyword arguments for pipe_base Task.\_\_init\_\_
         """
         RefMatchTask.__init__(self, refObjLoader, schema=schema, **kwargs)
+
+        if self.config.doWriteOutput and not schema is None:
```
parejkoj: `schema is not None`

python/lsst/meas/astrom/astrometry.py
```diff
@@ -269,6 +283,9 @@ def solve(self, exposure, sourceCat):
             (iterNum, len(tryRes.matches), tryMatchDist.distMean.asArcseconds(),
                 tryMatchDist.distStdDev.asArcseconds()))
 
+        for m in res.matches:
+            if self.usedKey:
```
parejkoj: probably should be explicit here: `if self.usedKey is not None`

python/lsst/meas/astrom/astrometry.py
```diff
@@ -147,6 +154,13 @@ def __init__(self, refObjLoader, schema=None, **kwargs):
         @param[in] kwargs  additional keyword arguments for pipe_base Task.\_\_init\_\_
         """
         RefMatchTask.__init__(self, refObjLoader, schema=schema, **kwargs)
+
+        if self.config.doWriteOutput and not schema is None:
+            self.usedKey = schema.addField("calib_astrometryUsed", type="Flag",
+                                               doc="set if source was used in astrometric calibration")
```
parejkoj: funky indentation

parejkoj:
And we need a unittest of this functionality here as well.

python/lsst/meas/astrom/astrometry.py
pgee2000: I would like to leave this consistent with the other calib modules.

python/lsst/meas/astrom/astrometry.py
pgee2000: I'm not a fan of this code standard, but OK.

python/lsst/meas/astrom/astrometry.py
pgee2000: OK

python/lsst/meas/astrom/astrometry.py
pgee2000: OK

pgee2000:
I have not bothered to add a unit test here.  I don't think it is worth the bother, and writing any kind of intelligent unit test requires an understanding of the Task, which is well beyond the scope of this ticket.  I did add one to photoCal, where it makes more sense.

python/lsst/meas/astrom/astrometry.py
PaulPrice: I agree that this should be changed. This `Task` does not do any writing, so the name is misleading.

PaulPrice:
It looks like you've merged prematurely.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -22,6 +22,8 @@
 from __future__ import absolute_import, division, print_function
 from builtins import range
 
+import random
+
```
laurenam: It doesn't look like this is used.

PR #69, opened 2017-05-03
DM-10453: Fix bugs in matchPessimisticB

morriscb:
The major bug fixed in this commit involves how the lookup table _pair_id_array in pessimistic_pattern_matcher_b_3D.py was filled. Previous to this commit the vertical slices of this matrix (line 130) was incorrectly filled with the second id of the reference pair. This mean that the lower half of the matrix contained incorret data and the last row slice (_pair_id_array[-1, :]) contained only one value of n_reference-1. This array is imporant as it is used to retrieve data needed to compare spokes in our pattern. This fix now fills the lower triangle of the matrix with the correct data for later lookup and comparison.

matchPessimisticB.py: Fixed lastMatchedPatterna and failedPatternList bugs. Changed default from list type to None type.
pessimistic_pattern_matcher_b_3D.py: Fixed pair_id data structure bug and changed output structs to not fill with partial results.

Removed default empty string from MatchPessimisticTolerance.

matchPessimisticB.py: Changed default input from list to None type.

### DIFF
--- a/python/lsst/meas/astrom/matchPessimisticB.py
+++ b/python/lsst/meas/astrom/matchPessimisticB.py
@@ -62,7 +62,8 @@ class MatchTolerancePessimistic(MatchTolerance):
         self.autoMaxMatchDist = autoMaxMatchDist
         self.maxShift = maxShift
         self.lastMatchedPattern = lastMatchedPattern
-        self.failedPatternList = failedPatternList
+        if failedPatternList is None:
+            self.failedPatternList = []
 
 
 class MatchPessimisticBConfig(pexConfig.Config):
@@ -451,22 +452,24 @@ class MatchPessimisticBTask(pipeBase.Task):
                     pattern_skip_array=np.array(
                         match_tolerance.failedPatternList)
                 )
-            if len(matcher_struct.matches) < 0 and \
-               match_tolerance.lastPattern is None and \
-               try_idx == 0:
-                # If we found a pattern on a previous run and can't
-                # find an optimistic match with the harshest tolerances,
-                # the match we found previously was likely bad. We append
-                # its index to a list so we do not use it in future match
-                # iterations.
+            if try_idx == 0 and \
+               len(matcher_struct.matches) == 0 and \
+               match_tolerance.lastMatchedPattern is not None:
+                # If we found a pattern on a previous match-fit iteration and
+                # can't find an optimistic match on our first try with the 
+                # tolerances as found in the previous match-fit,
+                # the match we found in the last iteration was likely bad. We
+                # append the bad match's index to the a list of
+                # patterns/matches to skip on subsequent iterations.
                 match_tolerance.failedPatternList.append(
-                    match_tolerance.lastPattern)
+                    match_tolerance.lastMatchedPattern)
+                match_tolerance.lastMatchedPattern = None
             elif len(matcher_struct.matches) > 0:
                 # Match found, save a bit a state regarding this pattern
                 # in the struct and exit.
                 match_tolerance.maxShift = afwgeom.Angle(matcher_struct.shift,
                                                          afwgeom.arcseconds)
-                match_tolerance.lastPattern = matcher_struct.pattern_idx
+                match_tolerance.lastMatchedPattern = matcher_struct.pattern_idx
                 break
 
         # A match has been found, return our list of matches and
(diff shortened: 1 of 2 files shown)


python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -451,22 +452,23 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
                     pattern_skip_array=np.array(
                         match_tolerance.failedPatternList)
                 )
-            if len(matcher_struct.matches) < 0 and \
```
r-owen: Sorry I let that through earlier.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -451,22 +452,23 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
-               match_tolerance.lastPattern is None and \
-               try_idx == 0:
+            if try_idx == 0 and \
+               len(matcher_struct.matches) == 0 and \
+               match_tolerance.lastMatchedPattern is not None:
                 # If we found a pattern on a previous run and can't
                 # find an optimistic match with the harshest tolerances,
                 # the match we found previously was likely bad. We append
                 # its index to a list so we do not use it in future match
                 # iterations.
```
r-owen: These comments sound really helpful but I confess I don't see how they relate to the tests above. Can you think of some way to rephrase them a bit that would help make the connection clearer?

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -123,11 +123,11 @@ class we use id to reference the position in the input reference
             self._pair_id_array[ref_id, ref_id:] = sub_id_array[:, 1]
             self._pair_delta_array[ref_id, ref_id:, :] = sub_delta_array
             self._pair_dist_array[ref_id, ref_id:] = sub_dist_array
-
+            
             # Don't fill the array column wise if we are on the last object
             # to avoid array overrun.
```
r-owen: Could you clarify this comment? I think you fill the array if we are not on the last object, but if you are on the last object then you don't fill the array at all. If that's really what you are doing then something like this might be clearer:

    # If not the last object, append ?? to the _pair arrays. Otherwise do...??? to avoid array overrun

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -822,13 +823,17 @@ def _create_pattern_spokes(self, src_ctr, src_delta_array, src_dist_array,
...
-                break
+            if len(ref_spoke_list) >= n_match - 2:
+                # Set the struct data and return the struct.
+                output_spokes.ref_spoke_list = ref_spoke_list
+                output_spokes.src_spoke_list = src_spoke_list
+                return output_spokes
+
         return output_spokes
 
     # TODO: Possibly clean up the arguments here by pre computing
```
r-owen: I realize this is pre-existing code but...can you please file a ticket for this cleanup and reference the ticket in the TODO code comment? That's how we are handling TODOs these days in order to keep track of them.

morriscb: This was left over when when I was thinking about changing the function inputs. @ctslater and I had a look and tried a marginally simpler version but we both decided it wasn't really any better. I have thus removed the TODO statement.

python/lsst/meas/astrom/matchPessimisticB.py
morriscb: Let me know if these comments are better.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -455,11 +455,12 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
-                # find an optimistic match with the harshest tolerances,
-                # the match we found previously was likely bad. We append
-                # its index to a list so we do not use it in future match
-                # iterations.
+                # If we found a pattern on a previous match-fit interation and
+                # can't find an optimistic match on our first try with the 
+                # tolerances as found in the previous match-fit,
+                # the match we found in the last interation was likely bad. We
+                # append the bad match's index to the a list of
+                # patterns/matches to skip on subsiquent iterations.
```
r-owen: That's much clearer. Thanks! Two minor typos: "interation" -> "iteration" (twice) and "subsiquent" -> "subsequent".

PR #71, opened 2017-05-08
Add calib_astrometryUsed flag to AstrometryTask

pgee2000:
Previous merge was accidental.  The actual changes are here.
Includes a quick test which only shows that the task is able
to write to the schema and sources table.

### DIFF
--- a/tests/testAstrometryTask.py
+++ b/tests/testAstrometryTask.py
@@ -85,6 +85,46 @@ class TestAstrometricSolver(lsst.utils.tests.TestCase):
         """
         self.doTest(afwGeom.RadialXYTransform([0, 1.01, 1e-7]))
 
+    def testUsedFlag(self):
+        """Test that the solver will record number of sources used to table
+           if it is passed a schema on initialization.
+        """
+        distortedWcs = afwImage.DistortedTanWcs(self.tanWcs, afwGeom.IdentityXYTransform())
+        self.exposure.setWcs(distortedWcs)
+        loadRes = self.refObjLoader.loadPixelBox(bbox=self.bbox, wcs=distortedWcs, filterName="r")
+        refCat = loadRes.refCat
+        refCentroidKey = afwTable.Point2DKey(refCat.schema["centroid"])
+        refFluxRKey = refCat.schema["r_flux"].asKey()
+
+        sourceSchema = afwTable.SourceTable.makeMinimalSchema()
+        measBase.SingleFrameMeasurementTask(schema=sourceSchema)  # expand the schema
+        config = AstrometryTask.ConfigClass()
+        config.wcsFitter.order = 2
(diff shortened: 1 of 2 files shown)


laurenam:
I still don't see where the `import random` in astrometry.py is being used.

tests/testAstrometryTask.py
```diff
@@ -85,6 +85,47 @@ def testRadial(self):
...
+        refCat = loadRes.refCat
+        refCentroidKey = afwTable.Point2DKey(refCat.schema["centroid"])
+        refFluxRKey = refCat.schema["r_flux"].asKey()
+
+        sourceSchema = afwTable.SourceTable.makeMinimalSchema()
+        measBase.SingleFrameMeasurementTask(schema=sourceSchema)  # expand the schema
+        config = AstrometryTask.ConfigClass()
+        config.wcsFitter.order = 2
+        config.wcsFitter.numRejIter = 0
+        #   schema must be passed to the solver task constructor
```
parejkoj: only one space after hash.

tests/testAstrometryTask.py
```diff
@@ -85,6 +85,47 @@ def testRadial(self):
...
+            src = sourceCat.addNew()
+            src.set(sourceCentroidKey, refObj.get(refCentroidKey))
+            src.set(sourceFluxKey, refObj.get(refFluxRKey))
+            src.set(sourceFluxSigmaKey, refObj.get(refFluxRKey)/100)
+
+        results = solver.run(
+            sourceCat=sourceCat,
+            exposure=self.exposure,
+        )
+        #   check that the used flag is set the right number of times
```
parejkoj: only one space after has.

tests/testAstrometryTask.py
```diff
@@ -85,6 +85,47 @@ def testRadial(self):
...
+        count = 0
+        flagKey = sourceCat.getSchema().find("calib_astrometryUsed").key
+        for source in sourceCat:
```
parejkoj: I think this for loop can be just `count = (sourceCat['calib_astrometryUsed']==True).sum()`

pgee2000: I think that I originally I tried this, or a slight variant, as I did in the photCal test.  But I got 'Record data is not contiguous in memory. ' error.   I did not stop to explore it, but just changed it to a loop.
Let me know if you think it matters.  Since I built this out of a test which already existed, I didn't particularly want to mess around with it.

parejkoj: I'm surprised that it won't work in this particular context, but I haven't tried. Since it's in the test code, I guess it's fine.

pgee2000:
I also removed the import random as per Lauren comment.  It was used by the reserve candidates code, which was later removed.

PR #72, opened 2017-05-29
DM-9751: Verify the performance of new matchPessimisticB code on selected test fields

morriscb:
Several stability improvements also made to matcher during this verification.

matchPessimisticB.py: Reordered cut calculations for clarity. Changed first matcher iteration to only use optimistic mode only after a match has been found on a previous match/fit iteration. Changed cut on match distance to be a 2 sigma clip instead of a hard cut.
pessimistic_pattern_matcher_b_3D.py: Added least squares fit to shift and rotation matrix in intermediate verify. removed distance cut from final verify. This cut is now done in matchPessimisticB.

Removed max_dist from _match_sources method.

Linter pass and editted sigma clip start.

### DIFF
--- a/python/lsst/meas/astrom/matchPessimisticB.py
+++ b/python/lsst/meas/astrom/matchPessimisticB.py
@@ -36,8 +36,8 @@ class MatchTolerancePessimistic(MatchTolerance):
         """Construct a MatchPessimisticTolerance
 
         MatchPessimisticBTask relies on several state variables to be
-        preserved over different iterations in the matchAndFitWcs loop of
-        AstrometryTask.
+        preserved over different iterations in the
+        AstrometryTask.matchAndFitWcs loop of AstrometryTask.
 
         Parameters
         ----------
@@ -54,7 +54,7 @@ class MatchTolerancePessimistic(MatchTolerance):
             a successful pattern match was found.
         failedPatternList : list of ints
             List of ints specifying indicies in the magnitude sourced source
-            array to skip. These are skipped are pervious iterations that are
+            array to skip. These are skipped are previous iterations that are
             likely false positives due to the code having to soften after a
             pattern is matched.
         """
@@ -135,11 +135,11 @@ class MatchPessimisticBConfig(pexConfig.Config):
         doc="How to select sources for cross-matching. The default "
             "matcherSourceSelector removes objects with low S/N, bad "
             "saturated objects, edge objects, and interpolated objects.",
-        default="matcher"
+        default="matcherPessimistic"
     )
 
     def setDefaults(self):
-        sourceSelector = self.sourceSelector["matcher"]
+        sourceSelector = self.sourceSelector["matcherPessimistic"]
         sourceSelector.setDefaults()
 
     def validate(self):
@@ -342,18 +342,6 @@ class MatchPessimisticBTask(pipeBase.Task):
             lsst.afw.table.ReferenceMatch, a MatchTolerance object
         """
 
-        # Set configurable defaults when we encounter None type or set
-        # state based on pervious run of matchandfit.
-        if match_tolerance.maxShift is None:
-            maxShiftArcseconds = (self.config.maxOffsetPix *
-                                  wcs.pixelScale().asArcseconds())
-        else:
-            # We don't want to clamp down to hard on the allowed shift so
-            # we test that the smallest we ever allow is the pixel scale.
-            maxShiftArcseconds = np.max(
-                (match_tolerance.maxShift.asArcseconds(),
-                 wcs.pixelScale().asArcseconds()))
-
         # Load the source and reference catalog as spherical points
         # in numpy array. We do this rather than relying on internal
         # lsst C objects for simplicity and because we require
@@ -376,6 +364,18 @@ class MatchPessimisticBTask(pipeBase.Task):
             flux = srcObj.getPsfFlux()
             src_array[src_idx, :] = \
                 self._latlong_flux_to_xyz_mag(theta, phi, flux)
+
+        # Set configurable defaults when we encounter None type or set
+        # state based on previous run of AstrometryTask._matchAndFitWcs.
+        if match_tolerance.maxShift is None:
+            maxShiftArcseconds = (self.config.maxOffsetPix *
+                                  wcs.pixelScale().asArcseconds())
+        else:
+            # We don't want to clamp down too hard on the allowed shift so
+            # we test that the smallest we ever allow is the pixel scale.
+            maxShiftArcseconds = np.max(
+                (match_tolerance.maxShift.asArcseconds(),
+                 wcs.pixelScale().asArcseconds()))
 
         # If our tolerances are not set from a previous run, estiamte a
         # starting tolerance guess from statistics of patterns we can create
@@ -402,23 +402,28 @@ class MatchPessimisticBTask(pipeBase.Task):
         # the consensus mode of the matcher. If not default to Optimistic
         # pattern matcher behavior.
         numConsensus = self.config.numPatternConsensus
-        if ref_array.shape[0] < self.config.numBrightStars or \
-           src_array.shape[0] < self.config.numBrightStars:
+        minObjectsForConsensus = \
+            self.config.numBrightStars + self.config.numPointsForShapeAttempt
+        if ref_array.shape[0] < minObjectsForConsensus or \
+           src_array.shape[0] < minObjectsForConsensus:
             numConsensus = 1
 
-        self.log.debug("Current tol maxDist: %.4f..." % maxMatchDistArcSec)
-        self.log.debug("Current shift: %.4f" % maxShiftArcseconds)
-        self.log.debug("Current rot: %.4f" % self.config.maxRotationDeg)
+        self.log.debug("Current tol maxDist: %.4f arcsec" %
+                       maxMatchDistArcSec)
+        self.log.debug("Current shift: %.4f arcsec" %
+                       maxShiftArcseconds)
 
         # Create our matcher object.
         pyPPMb = PessimisticPatternMatcherB(ref_array[:, :3], self.log)
 
         # Start the ineration over our tolerances.
         for try_idx in range(self.config.matcherIterations):
-            if try_idx == 0:
+            if try_idx == 0 and \
+               match_tolerance.lastMatchedPattern is not None:
                 # If we are on the first, most stringent tolerance,
-                # the matcher should behave like an optimistic pattern
-                # matcher. Exiting at the first match.
+                # and have already found a match, the matcher should behave
+                # like an optimistic pattern matcher. Exiting at the first
+                # match.
                 matcher_struct = pyPPMb.match(
                     source_array=src_array,
                     n_check=self.config.numPointsForShapeAttempt,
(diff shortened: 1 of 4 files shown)


python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -472,23 +476,82 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
-            match.distance = match.first.getCoord().angularSeparation(
-                match.second.getCoord())
-            matches.append(match)
+                                   distances):
+            if dist < dist_cut:
+                match = afwTable.ReferenceMatch()
+                match.first = refCat[match_ids[1]]
+                match.second = sourceCat[match_ids[0]]
+                match.distance = match.first.getCoord().angularSeparation(
+                    match.second.getCoord())
```
r-owen: Why can't you use the distance you already know? If you can't, then a code comment explaining would be helpful.

Also I find it very confusing to have `matches` use our usual definition of a list of `ReferenceMatch` here, but `matcher_struct.matches` mean something entirely different. For now I suggest renaming `matcher_struct.matches` to something else, such as `matched_ids`. But I hope in the long run `matcher_struct` can contain a normal match list (list of `ReferenceMatch`) and get rid of the separate distance field.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -472,23 +476,82 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
+        Find the distance cut by recursively cliping by n_sigma.
+
+        The algorithm converges once the length of the array on the
+        nth iteration is equal to the the length of the array on the
+        n+1th interation.
+
+        Parameters
+        ----------
+        dists : float array
+            float array distances to caculate the cut on.
```
r-owen: Please consider `values`; it could be distances, but it could be anything. If you use distances then the next obvious question users will ask is "what units" (that was my first reaction until I realized it didn't matter).

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -472,23 +476,82 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
+            hold_cut = np.max(dists)
+        else:
+            hold_cut = input_cut
+        # Start clipping
+        for clip_idx in xrange(100):
+            # Find the mask and current total objects.
+            mask = (dists < hold_cut)
+            total = len(dists[mask])
+            # Test to see if the current total is equal to the previous.
+            # If it is we exit returning the pervious cut.
```
r-owen: pervious -> previous (perhaps a global search?)

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -2,11 +2,49 @@
...
+
+    Given a flattened rotation matrix, a N point pattern from the source
+    catalog and the reference pattern the sources match to, compute return
+    the squared differences between the points in the two patterns after
+    rotation.
+
+    Parameters
+    ----------
+    flattened_rot_matrix : float array
+        A flattened array representing a 3x3 rotation matrix.
```
r-owen: I suggest you use a normal 2D array, instead. It makes the API easier to understand, and the matrix needs to be unflattened at some point -- might as well do it before calling this.

if you insist on the data being a 1-D array then you must explain the order in which the elements are stored.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -2,11 +2,49 @@
...
 from scipy.spatial import cKDTree
 
 import lsst.pipe.base as pipeBase
 
 
+def _rotation_matrix_chi_sq(flattened_rot_matrix, src_pattern, ref_pattern):
+    """Compute the squared differences for least squares fitting.
+
+    Given a flattened rotation matrix, a N point pattern from the source
+    catalog and the reference pattern the sources match to, compute return
```
r-owen: "compute return" -> "compute"

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -2,11 +2,49 @@
...
+        A array containing N, 3 vectors
+    ref_pattern : float array of 3 vectors
+        A array containing N, 3 vectors
+
+    Returns
+    -------
+    float array
+        Array of differences between the points of the source pattern
+        rotated into the reference frame and the reference pattern rotated
+        into the source frame.
```
r-owen: I'm afraid I don't know what this means. What is the "difference between the points..."?

Also, why do you rotate the source into the reference frame *and also* rotate the reference pattern into the source frame. Naively I would expect only one rotation would be required to compare patterns of points in the two frames.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -123,9 +161,9 @@ class we use id to reference the position in the input reference
             self._pair_id_array[ref_id, ref_id:] = sub_id_array[:, 1]
             self._pair_delta_array[ref_id, ref_id:, :] = sub_delta_array
             self._pair_dist_array[ref_id, ref_id:] = sub_dist_array
-            
+
             # Don't fill the array column wise if we are on the last as
-            # these pairs have already been filled by pervious 
+            # these pairs have already been filled by pervious
```
r-owen: pervious -> previous

tests/test_MatchPessimisticB.py
```diff
@@ -157,7 +152,7 @@ def plot(catalog, symbol):
 
             # Current skip refObj 15 as it matches to source 12.
             # All other matches are correct.
```
r-owen: It appears this comment is outdated, based on your change below

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -402,8 +402,10 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
         # the consensus mode of the matcher. If not default to Optimistic
         # pattern matcher behavior.
         numConsensus = self.config.numPatternConsensus
-        if ref_array.shape[0] < self.config.numBrightStars or \
-           src_array.shape[0] < self.config.numBrightStars:
+        if ref_array.shape[0] < self.config.numBrightStars + \
+           self.config.numPointsForShapeAttempt or \
+           src_array.shape[0] < self.config.numBrightStars + \
+           self.config.numPointsForShapeAttempt:
```
morriscb: Should I make this intermediate variable or would you recommend putting it in the config?

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: The matrix needs to be flattened for the fitter that is used later in the code (scipy.optimize.leastsq) which only works with flattened arrays as input (I checked). I added an extra line of explanation to the doc string.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Added a extra set of explanation to this.  The reason the second rotation exist is to test that the fitted rotation is a valid rotation that is np.dot(matrix.transpose(), matrix) = identity. I could equivalently test the above holds instead of doing the inverse rotation on the reference pattern. This may be clearer. This would replace line 45 with diff_ident = np.dot(rot_matrix.transpose(), rot_matrix) - np.identity(3).

python/lsst/meas/astrom/matchPessimisticB.py
r-owen: I suggest an intermediate variable since it is a simple sum of two config variables. The only reason to put it in the config would be if the user might want to use some other value, but unless that's likely it just adds clutter.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
r-owen: If it would be clearer and no less robust it sounds like a good change

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -469,26 +475,102 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
-            matches.append(match)
+        for match_id_pair, dist_arcsec in zip(matcher_struct.match_ids,
+                                              distances_arcsec):
+            if dist_arcsec < dist_cut_arcsec:
+                match = afwTable.ReferenceMatch()
+                match.first = refCat[match_id_pair[1]]
+                match.second = sourceCat[match_id_pair[0]]
+                # We compute the true distance along and sphere instead
+                # and store it as an afw.geom.Angle object. The previous
+                # distances we used were aproximate.
```
r-owen: Unfortunately you can't store the distance as an `Angle` because `ReferenceMath.distance` is a C++ double. The angle is almost certainly being silently cast to radians; please make the cast explicit (and update the code comment), e.g.:
```
match.distance = match.first.getCoord().angularSeparation(match.second.getCoord()).asRadians()
```

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -2,11 +2,53 @@
...
+    Given a flattened rotation matrix, a N point pattern from the source
+    catalog and the reference pattern the sources match to, compute
+    the squared differences between the points in the two patterns after
+    rotation.
+
+    Parameters
+    ----------
+    flattened_rot_matrix : float array
+        A flattened array representing a 3x3 rotation matrix. The array is
+        flattened to comply with the API of scipy.optimize.leastsq
```
r-owen: Please document the oder of the data -- how do the 9 values in the array represent the 3x3 values in the matrix? Is it elements [0,0], [0,1], [0,2], [1,0], ... or the transpose or ...?

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -227,7 +271,7 @@ def match(self, source_array, n_check, n_match, n_agree,
 
         # Initialize output struct.
         output_match_struct = pipeBase.Struct(
-            matches=[],
+            match_ids=[],
```
r-owen: Thanks!

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -1023,23 +1073,74 @@ def _intermediate_verify(self, src_pattern, ref_pattern, shift_rot_matrix,
 
         Returns
         -------
-        bool
-           Return true if all of the points in our source pattern are within
-           max_dist_rad of their matched reference objects.
+        float array
```
r-owen: `float array or None`

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -402,8 +402,12 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
         # pattern matcher behavior.
         numConsensus = self.config.numPatternConsensus
-        if ref_array.shape[0] < self.config.numBrightStars or \
-           src_array.shape[0] < self.config.numBrightStars:
+        minRefAndSrcObjectsForConsensus = \
+            self.config.numBrightStars + self.config.numPointsForShapeAttempt
+        if ref_array.shape[0] < self.config.numBrightStars + \
+           self.config.numPointsForShapeAttempt or \
+           src_array.shape[0] < self.config.numBrightStars + \
+           self.config.numPointsForShapeAttempt:
```
r-owen: Thanks for defining `minRefAndSrcObjectsForConsensus` but then please use it in the test. (flake8 should have warned you that the variable is unused?) Also, consider shortening the name to `minObjectsForConsensus`.

PR #73, opened 2017-06-14
DM-10931: Removed print Fix variable name bug and remove print statements in matchPessimisticB. and fixed distances_rad name.

morriscb:


python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -273,7 +273,7 @@ def match(self, source_array, n_check, n_match, n_agree,
         # Initialize output struct.
         output_match_struct = pipeBase.Struct(
             match_ids=[],
-            distances=[],
+            distances_rad=[],
```
r-owen: Also update the numpy docs above to match: the output is `distances_rad` instead of `distances`

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -505,7 +503,6 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
         dist_cut_arcsec = np.max(
             (2 * wcs.pixelScale().asArcseconds(),
              clipped_dist_arcsec, maxMatchDistArcSec))
-        print("Dist cut", dist_cut_arcsec, "arcsec")
```
r-owen: If you think any of these print statements may be of interest to somebody then I strongly suggest you make these log statements at some suitable verbosity. INFO is usually printed. TRACE is the next step in verbosity, then DEBUG. So taking a guess at TRACE (and providing the data as a separate argument so the mess is only formatted if it is to be printed -- though the computation of `max` will presumably be more expensive than the formatting):

    self.log.trace("max distance = %s arcsec", distances_arcsec.max())

morriscb: These where more for me to check what was going on in the unittest when I was trying to update the code. The useful information about matching is already stored in the debug logger.

PR #75, opened 2017-08-14
DM-5922: Rework camera geometry to use the replacement for XYTransform

r-owen:


tests/testFitTanSipWcsHighOrder.py
```diff
@@ -113,7 +113,7 @@ def plotWcs(self, wcs0, wcs1, bbox, xyTransform):
                 pixelPos0 = afwGeom.Point2D(x, y)
                 skyCoord = wcs0.pixelToSky(pixelPos0)
                 pixelPos1 = wcs1.skyToPixel(skyCoord)
-                distortedPos = xyTransform.forwardTransform(pixelPos0)
+                distortedPos = xyTransform.applyForward(pixelPos0)
```
TallJimbo: Might be appropriate to change the name of the `xyTransform` argument/variable here as well.

PR #76, opened 2017-10-24
DM-12207: Add color selection for photocal

PaulPrice:


python/lsst/meas/astrom/directMatch.py
```diff
@@ -88,11 +95,22 @@ def run(self, catalog, filterName=None):
...
+            return Struct(matches=[], matchMeta=matchMeta)
         refData = self.refObjLoader.loadSkyCircle(circle.center, circle.radius, filterName)
-        matches = afwTable.matchRaDec(refData.refCat, catalog, self.config.matchRadius*arcseconds)
-        self.log.info("Matched %d from %d input and %d reference sources" %
-                      (len(matches), len(catalog), len(refData.refCat)))
-        return Struct(matches=matches, matchMeta=matchMeta)
+        refCat = refData.refCat
+        refSelection = self.referenceSelection.selectSources(refCat)
+        if len(refSelection.sourceCat) == 0:
+            self.log.warn("No objects selected from %d objects in reference catalog", len(refCat))
```
parejkoj: Why run matches after this, if there are no ref sources? Shouldn't you also short-circuit an empty `return`?

PR #79, opened 2017-12-22
tickets/DM-12431: re-Verify performance of matchPessimisticB with new distortions.

morriscb:
Changed behavior of final output matches to the smaller of
a sigma clipped cut and the current maxMatchDistance.

This change adds a bit more stability to output match quality and fit/match cycle decent.

### DIFF
--- a/python/lsst/meas/astrom/matchPessimisticB.py
+++ b/python/lsst/meas/astrom/matchPessimisticB.py
@@ -501,7 +501,11 @@ class MatchPessimisticBTask(pipeBase.Task):
         # the reference and source catalog. For the current astrometric solver
         # we need to remove as many false positives as possible before sending
         # the matches off to the solver. The low value of 100 and high value of
-        # 2 are the low number of sigma and high respectively.
+        # 2 are the low number of sigma and high respectively. The exact values
+        # were found after testing on data of various reference/source
+        # densities and astrometric distortion quality, specifically the
+        # visits:  HSC (3358), DECam (406285, 410827),
+        # CFHT (793169, 896070, 980526).
         distances_arcsec = np.degrees(matcher_struct.distances_rad) * 3600
         clip_max_dist = np.max(
             (sigmaclip(distances_arcsec, low=100, high=2)[-1],
--- a/python/lsst/meas/astrom/matchPessimisticB.py
+++ b/python/lsst/meas/astrom/matchPessimisticB.py
@@ -437,7 +437,7 @@ class MatchPessimisticBTask(pipeBase.Task):
         # Start the iteration over our tolerances.
         for soften_pattern in range(self.config.matcherIterations):
             for soften_dist in range(self.config.matcherIterations):
-                if try_idx == 0 and soften_idx == 0 and \
+                if soften_pattern == 0 and soften_dist == 0 and \
                     match_tolerance.lastMatchedPattern is not None:
                     # If we are on the first, most stringent tolerance,
                     # and have already found a match, the matcher should behave
--- a/python/lsst/meas/astrom/matchPessimisticB.py
+++ b/python/lsst/meas/astrom/matchPessimisticB.py
@@ -435,8 +435,8 @@ class MatchPessimisticBTask(pipeBase.Task):
 
         match_found = False
         # Start the iteration over our tolerances.
-        for try_idx in range(self.config.matcherIterations):
-            for soften_idx in range(self.config.matcherIterations):
+        for soften_pattern in range(self.config.matcherIterations):
+            for soften_dist in range(self.config.matcherIterations):
                 if try_idx == 0 and soften_idx == 0 and \
                     match_tolerance.lastMatchedPattern is not None:
                     # If we are on the first, most stringent tolerance,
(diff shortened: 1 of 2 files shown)


python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -490,15 +491,16 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
         # We pick the largest of the the unsoftened maxMatchDistArcSec or
         # a user specified distance in pixels. This prevents the
         # AstrometryTask._matchAndFitWCS from over-fitting to a small number of
         # objects and also allows the WCS fitter to bring in more matches as
         # the WCS fit improves.
-        dist_cut_arcsec = np.max(
-            (self.config.minMatchDistPixels * wcs.pixelScale().asArcseconds(),
-             maxMatchDistArcSec)
-        )
+        dist_cut_arcsec = np.min((clip_max_dist, maxMatchDistArcSec))
```
jdswinbank: Seems like this makes the comment above untrue, right? You are no longer picking the largest of anything, but the smallest of the sigma clipped cut and `maxMatchDistArcSec`.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -500,7 +512,15 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
         # AstrometryTask._matchAndFitWCS from over-fitting to a small number of
         # objects and also allows the WCS fitter to bring in more matches as
         # the WCS fit improves.
-        dist_cut_arcsec = np.min((clip_max_dist, maxMatchDistArcSec))
+        if not np.isfinite(clip_max_dist):
+            clip_max_dist = maxMatchDistArcSec
+
```
jdswinbank: There are eight ` ` characters here that you don't want.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -432,73 +433,94 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
+                matches=[],
+                match_tolerance=match_tolerance,
+            )
+
         # The matcher returns all the nearest neighbors that agree between
         # the reference and source catalog. For the current astrometric solver
         # we need to remove as many false positives as possible before sending
         # the matches off to the solver.
-        distances_arcsec = np.degrees(matcher_struct.distances_rad) * 3600
+        distances_arcsec = np.degrees(matcher_struct.distances_rad) * 3600
```
jdswinbank: You added extra whitespace that you don't want to this line.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -432,73 +433,94 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
-        )
+        if not np.isfinite(clip_max_dist):
+            clip_max_dist = maxMatchDistArcSec
+        
+        if clip_max_dist < maxMatchDistArcSec and \
+           len(distances_arcsec[distances_arcsec < clip_max_dist]) < \
+           minMatchedPairs:
+            dist_cut_arcsec = maxMatchDistArcSec
+        else:
+            dist_cut_arcsec = np.min((clip_max_dist, maxMatchDistArcSec))
```
jdswinbank: These changes make the comment above untrue — you are not picking the largest of anything, but the minimum of `clip_max_dist` and `maxMatchDistArcSec` (except when you aren't).

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -432,73 +433,94 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
+            )
+
         # The matcher returns all the nearest neighbors that agree between
         # the reference and source catalog. For the current astrometric solver
         # we need to remove as many false positives as possible before sending
         # the matches off to the solver.
-        distances_arcsec = np.degrees(matcher_struct.distances_rad) * 3600
+        distances_arcsec = np.degrees(matcher_struct.distances_rad) * 3600 
+        clip_max_dist = np.max(
+            (sigmaclip(distances_arcsec, low=100, high=2)[-1],
```
jdswinbank: What's the justification for these values of `low` and `high`? They just seem like magic numbers here.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -432,73 +433,94 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
+                    # tolerances as found in the previous match-fit,
+                    # the match we found in the last iteration was likely bad. We
+                    # append the bad match's index to the a list of
+                    # patterns/matches to skip on subsequent iterations.
+                    match_tolerance.failedPatternList.append(
+                        match_tolerance.lastMatchedPattern)
+                    match_tolerance.lastMatchedPattern = None
+                    maxShiftArcseconds = \
+                        self.config.maxOffsetPix * wcs.pixelScale().asArcseconds()
+                    continue
```
jdswinbank: What does this `continue` buy you?

jdswinbank:
Seems like the second commit ought to be a couple of separate commits really.

python/lsst/meas/astrom/matchPessimisticB.py
```diff
@@ -432,73 +433,94 @@ def _doMatch(self, refCat, sourceCat, wcs, refFluxField, numUsableSources,
...
-                    self.config.maxOffsetPix * wcs.pixelScale().asArcseconds()
-                continue
-            elif len(matcher_struct.match_ids) > 0:
-                # Match found, save a bit a state regarding this pattern
-                # in the match tolerance class object and exit.
-                match_tolerance.maxShift = \
-                    matcher_struct.shift * afwgeom.arcseconds
-                match_tolerance.lastMatchedPattern = \
-                    matcher_struct.pattern_idx
+            for soften_idx in range(self.config.matcherIterations):
```
jdswinbank: Could you maybe clean up the language around here a bit please?

Specifically, both `soften_idx` and `try_idx` are "softening", they're just softening different things. Maybe clearer to use, say, `soften_pattern` and `soften_dist`, or something like that.

python/lsst/meas/astrom/matchPessimisticB.py
morriscb: You correct, it buys me nothing really. Just habit. It is now removed.

python/lsst/meas/astrom/matchPessimisticB.py
morriscb: The are the number of sigma for the clip. Low is 100 because the lower bound is always a value of 0 and high is a 2 sigma cut. I will add a comment explaining this. I could also make the high value a configurable if that is useful.

python/lsst/meas/astrom/matchPessimisticB.py
morriscb: The comment has been edited.

PR #81, opened 2018-03-16
DM-11162: Replace all use of Coord and subclasses with SpherePoint

r-owen:


tests/test_fitTanSipWcsHighOrder.py
```diff
@@ -39,25 +38,14 @@ class ApproximateWcsTestCase(lsst.utils.tests.TestCase):
...
-        metadata.set("CTYPE1", "RA---TAN")
-        metadata.set("CTYPE2", "DEC--TAN")
-        metadata.setDouble("CD1_1", 5.10808596133527E-05)
-        metadata.setDouble("CD1_2", 1.85579539217196E-07)
-        metadata.setDouble("CD2_2", -5.10281493481982E-05)
-        metadata.setDouble("CD2_1", -8.27440751733828E-07)
-        self.tanWcs = afwGeom.makeSkyWcs(metadata)
+        self.tanWcs = afwGeom.makeSkyWcs(crpix=self.crPix,
+                                         crval=afwGeom.SpherePoint(215.5, 53.0, afwGeom.degrees),
+                                         cdMatrix=afwGeom.makeCdMatrix(scale=5.1e-5*afwGeom.degrees,
```
rearmstr: Is there a need to simplify the cdMatrix here?

tests/test_transforms.py
```diff
@@ -525,11 +523,11 @@ def testFromMatches(self):
         initialIwcToSky = lsst.afw.geom.getIntermediateWorldCoordsToSky(initialWcs)
         for i in range(nPoints):
             refRec = ref.addNew()
-            skyPos = lsst.afw.geom.Point2D(
+            raDeg, decDeg = (
                 np.random.uniform(low=bbox.getMinX(), high=bbox.getMaxX()),
-                np.random.uniform(low=bbox.getMinY(), high=bbox.getMaxY())
+                np.random.uniform(low=bbox.getMinY(), high=bbox.getMaxY()),
```
rearmstr: Is this trailing comma needed?

tests/test_fitTanSipWcsHighOrder.py
r-owen: I don't understand the question. The primary changes is to use `makeSkyWcs(crpix, ...)` instead of `makeSkyWcs(metadata)`, saving the hassle of creating a bunch of FITS-WCS metadata. `makeSkyWcs(crpix, ...)` requires a CD matrix and `makeCdMatrix` is the easiest way I know of to make one.

tests/test_transforms.py
r-owen: It is not necessary but it is good Python style.

tests/test_fitTanSipWcsHighOrder.py
rearmstr: Yeah, sorry I wasn't explicit.  Your new cdMatrix has only a scale and rotation.  The previous cdMatrix looks like it is more complicated with some shear.

r-owen: Good point. I missed the difference between CD2_1 and CD1_2. I'll restore the original values. I've never seen a CD matrix like that before. I wonder if it was intentional.

PR #87, opened 2018-05-24
DM-9837 Update catalogStarSelector to new SourceSelector API

parejkoj:


python/lsst/meas/astrom/catalogStarSelector.py
```diff
@@ -141,18 +143,27 @@ def DebugInfo(name):
...
-        A list of PSF candidates may be used by a PSF fitter to construct a PSF.
-
-        @param[in] exposure  the exposure containing the sources
-        @param[in] sourceCat  catalog of sources that may be stars (an lsst.afw.table.SourceCatalog)
-        @param[in] matches  a match vector as produced by meas_astrom; required
-                            (defaults to None to match the StarSelector API and improve error handling)
-
-        @return an lsst.pipe.base.Struct containing:
-        - starCat  catalog of selected stars (a subset of sourceCat)
+    def selectSources(self, sourceCat, matches=None, exposure=None):
```
TallJimbo: Do we need to give `matches` a default value here?  (Or rather, is that what we usually do for `SourceSelectors` with `usesMatches = True`?)

parejkoj: I can only find two SourceSelectors that have `usesMatches = True`, so there really isn't a known behavior. The first thing it does is `if matches is None`, so I think it's fine to default it like that.

PR #88, opened 2018-06-01
DM-14628: meas_astrom pytest setup is missing E266

mrawls:


timj:
E266 is not one of the standard flake8 types we exclude. https://developer.lsst.io/python/style.html

jdswinbank:
...which (implicitly) suggests that you should this on a file-by-file basis rather than globally.

timj:
No. E266 has never been allowed. The problem is that `##` is sometimes used in code for doxygen. We therefore allow specific files to have E266 masked during the transition to numpydoc. Since `flake8` itself does not allow per file masking for the cases where the `##` is in use, we have to mask E266 for flake8 general command line usage. For pytest we try to mask the specific files until they can be fixed. We should not be globally masking E266 in pytest flake8. This is an issue with doxygen to numpydoc transition.

jdswinbank:
I'm struggling to parse the above (“E266 has never been allowed... we have to mask E266”), but I think the conclusion is (as above) that this is fine on a file-by-file basis, but not globally. Right?

setup.cfg
```diff
@@ -8,6 +8,6 @@ exclude =
 
 [tool:pytest]
 addopts = --flake8
-flake8-ignore = E133 E226 E228 N802 N803 N806
+flake8-ignore = E133 E226 E228 N802 N803 N806 E266
```
r-owen: [tool:pytest] allows us to exclude particular error codes for particular files; use use this, as per https://community.lsst.org/t/task-documentation-and-the-numpydoc-flake8-transitions e.g.

I believe the correct thing to add, instead of the E266 above, is the following after `tests/data/*`:

    astrometry.py E266
    fitTanSipWcs.py E266
    ref_match.py E266

which I figured out by removing the E266 exclusion from flake8 (earlier in setup.cfg) and then running flake8.

I'm glad you are fixing this -- I don't know how I missed it; for some reason scons tests works on my Mac and Jenkins without the explicit excludes, but I think they ought to be necessary.

timj:
I'll try again. E266 should never be ignored. For some specific cases where we have doxygen documentation with `##` we have to ignore it. We should ignore in the pytest section on a per-file basis so that we are clear which files are being masked. Since flake8 from the command line does not allow per file masking (or at least it didn't) you have to mask E266 everywhere. The pytest section is distinct. This will all get resolved when the doxygen is replaced with numpydoc.

setup.cfg
```diff
@@ -8,6 +8,6 @@ exclude =
...
     # TODO: remove E266 lines when Task documentation is converted to rst in DM-14207.
```
timj: This comment line is relevant (I think it was copied from another and left in by mistake) once the per-file ignores are added.

mrawls:
Thanks for the clarification, all. Should be all set now.

PR #116, opened 2019-03-06
DM-18314: reduce namespace confusion for matchOptimisticB

PaulPrice:
The 'matchOptimisticB' namespace is overloaded, which makes it
impossible to do:
    import lsst.meas.astrom.matchOptimisticB.matchOptimisticBContinued
    lsst.meas.astrom.matchOptimisticB.matchOptimisticBContinued.MatchOptimisticBTask
results in:
    AttributeError: 'builtin_function_or_method' object has no attribute 'matchOptimisticBContinued'

The 'matchOptimisticB' module contains a function called
'matchOptimisticB' which is being imported over the top of the
module, so rename that function. Move the MatchOptimisticBTask
from the 'matchOptimisticB' module so it can be accessed
without problems. Added a test that detects the problem.

### DIFF
--- a/python/lsst/meas/astrom/__init__.py
+++ b/python/lsst/meas/astrom/__init__.py
@@ -21,7 +21,7 @@
 #
 
 from .makeMatchStatistics import *
-from .matchOptimisticB import *
+from .matchOptimisticBTask import *
 from .polynomialTransform import *
 from .scaledPolynomialTransformFitter import *
 from .sipTransform import *
@@ -31,7 +31,6 @@ from . import sip
 from .ref_match import *
 from .astrometry import *
 from .approximateWcs import *
-from .matchOptimisticB import *
 from .matchPessimisticB import *
 from .setMatchDistance import *
 from .display import *
--- a/python/lsst/meas/astrom/matchOptimisticB/__init__.py
+++ b/python/lsst/meas/astrom/matchOptimisticB/__init__.py
@@ -1,3 +1,2 @@
 
 from .matchOptimisticB import *
-from .matchOptimisticBContinued import *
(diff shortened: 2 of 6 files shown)


python/lsst/meas/astrom/matchOptimisticB/matchOptimisticB.cc
```diff
@@ -102,7 +102,7 @@ PYBIND11_MODULE(matchOptimisticB, mod) {
             (ProxyVector(*)(afw::table::SimpleCatalog const &, afw::geom::SkyWcs const &)) & makeProxies,
             "posRefCat"_a, "tanWcs"_a);
 
-    mod.def("matchOptimisticB", &matchOptimisticB, "posRefCat"_a, "sourceCat"_a, "control"_a, "wcs"_a,
+    mod.def("matchCatalogsOptimisticB", &matchOptimisticB, "posRefCat"_a, "sourceCat"_a, "control"_a, "wcs"_a,
```
r-owen: I am puzzled why you are making this particular change. I _think_ it would be preferable to leave the name alone and stop hoisting the `matchOptimisticB` function up a level (meaning it can only be referenced by `lsst.meas.astrom.matchOptimisticB.matchOptimisticB`). The other public symbols can still be hoisted, as you are doing, leaving most code working without any changes to the C++.

I am concerned that renaming this function will be very confusing because it doesn't match the C++ function, nor the name of the pybind11 file, nor the name of the task. I really hope you can avoid this rename.

PaulPrice:
@r-owen: I've changed it so that the function name is preserved. I'm not completely happy with this solution, as the function is still being imported into the `lsst.meas.astrom` namespace (via the `matchOptimisticBTask`) and clobbering the module, but it seems that we're escaping the nasty effects of that by luck. Nevertheless, a test is now in place to ensure our luck continues.

r-owen:
If it works feel free to merge. But consider omitting "matchOptimisticB" from `__all__` in `matchOptimisticBContinued.py`. Then it wont be hoisted and I think we'd both be happier.

r-owen:
If it works feel free to merge. But consider omitting "matchOptimisticB" from __all__ in matchOptimisticBContinued.py. Then it wont be hoisted and I think we'd both be happier

PR #119, opened 2019-03-22
DM-18400: Number of matches is smaller than request

morriscb:


python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -314,31 +316,55 @@ def match(self, source_array, n_check, n_match, n_agree,
...
+            # survive still.
+            n_matched = np.sum(
+                match_sources_struct.distances_rad < max_dist_rad)
+            clipped_max_dist = np.max((
+                sigmaclip(match_sources_struct.distances_rad,
+                          low=100,
+                          high=2)[-1],
+                1e-16))
+            n_matched_clipped = np.sum(
+                match_sources_struct.distances_rad < clipped_max_dist)
```
ctslater: There seems to be repeated block of code between lines 324-336 and 354-365. This would be better if it was given some structure, e.g. pulled out into a function. These two blocks also repeat usage of the same variable names, which is both confusing and risky.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -314,31 +316,55 @@ def match(self, source_array, n_check, n_match, n_agree,
...
-                          max_dist_rad)
-                ).x.reshape((3, 3))
-                match_sources_struct = self._match_sources(
-                    source_array[:, :3], fit_shift_rot_matrix)
+            # The shift/rotation matrix returned by
+            # ``_construct_pattern_and_shift_rot_matrix``, above, was
+            # based on only six points. Here, we refine that result by
+            # using all of the good matches from the “final verification”
+            # step, above. This will produce a more consistent result.
+            fit_shift_rot_matrix = least_squares(
```
ctslater: This for loop (starting at line 257) is getting really long (> 100 lines) and the fitting stage starting here seems like a different stage compared to the earlier part of the loop; can it be extracted into a separate function? It would be nicer if this part had a name and defined inputs/outputs.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -1204,6 +1192,119 @@ def _test_rotation_agreement(self, rot_vects, max_dist_rad):
                 tot_consent += 1
         return tot_consent
 
+    def _final_verify(self,
+                      source_array,
+                      shift_rot_matrix,
+                      max_dist_rad,
+                      min_matches):
+        """Match the all sources into the reference cat using the shift/rot
```
ctslater: cat -> catalog

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -1204,6 +1192,119 @@ def _test_rotation_agreement(self, rot_vects, max_dist_rad):
...
+        matrix.
+
+        After the initial set of matches and found to have enough matches,
+        refit the shift/rot matrix for a more stable solution.
```
ctslater: "and found" probably isn't want you meant, but I'm not sure what you wanted. I don't know how to interpret the imperative phrasing; am I, the user, supposed to do the refitting or is that what this function does? This could use a clear statement of what the output is.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -1204,6 +1192,119 @@ def _test_rotation_agreement(self, rot_vects, max_dist_rad):
...
+              the shift/rotated frame. (`numpy.ndarray`, (N,)).
+            - ``max_dist_rad`` : Value of the max matched distance. Either
+              returning the input value of the 2 sigma clipped value of the
+              shift/rotated distances. (`float`).
+        """
+        output_struct = pipeBase.Struct(
+            match_ids=None,
+            distances_rad=None,
+            max_dist_rad=None,
+        )
```
ctslater: It would be simpler to return None on failure rather than a struct full of Nones. That way you wouldn't have to generate this at the start, the short-circuit error returns would be more obvious (`return None` is clearer than `return output_struct`), and the test in `match()` would be much simpler.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -1204,6 +1192,119 @@ def _test_rotation_agreement(self, rot_vects, max_dist_rad):
...
+
+        # Perform an iterative final verify.
+        match_sources_struct = self._match_sources(source_array,
+                                                   shift_rot_matrix)
+        cut_ids = match_sources_struct.match_ids[
+            match_sources_struct.distances_rad < max_dist_rad]
+
+        n_matched = len(cut_ids)
+        # Check clipped distances. The minimum value here
```
ctslater: Still need to extract the clipping calculation into a function rather than repeating it twice.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Pretty sure others had an opinion against just returning None when I first coded things this way. For consistency with the rest of the code I'm going to leave it as is.

python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
morriscb: Fixed

PR #120, opened 2019-03-27
DM-18643: Add sourceSelection and refSelection to AstrometryTask.solve()

erykoff:
Note that the default source and reference selectors (defined in
RefMatchConfig) do not down-select, so fixing this bug will not change the
default runs. However, anybody who thought they were doing source or
reference selection will see a change now that that is happening.


morriscb:
Eli, are you planning on closing this ticket?

PR #122, opened 2019-04-10
DM-18643: Move AstrometryTask source selection from "matcher" into AstrometryTask	

erykoff:
Implementing RFC-589, this moves the source selector into the astrometry task, and additionally fixes the reference selector so it is run in both matching and solving modes.


morriscb:
Looks good. I would suggest squashing the "Fix typo" commits.

PR #140, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


tests/test_astrometryTask.py
```diff
@@ -49,7 +49,7 @@ def setUp(self):
                                          cdMatrix=afwGeom.makeCdMatrix(scale=5.1e-5*lsst.geom.degrees))
         self.exposure = afwImage.ExposureF(self.bbox)
         self.exposure.setWcs(self.tanWcs)
-        self.exposure.setFilter(afwImage.Filter("r", True))
+        self.exposure.setFilterLabel(afwImage.FilterLabel(band="r", physical="rTest"))
```
kfindeisen, line 52: Do you actually want a physical filter here?

kfindeisen (approved):

tests/test_astrometryTask.py
parejkoj: Technically, no, since I made the code use `band`, but I would expect any data coming in to here to have both populated.

PR #141, opened 2021-01-08
DM-28258: Epoch is not getting passed consistently among reference catalog loading functions

laurenam:


parejkoj (approved):
This looks fine: `expMd` already has `epoch` as an `astropy.Time`.

It surprised me that AstrometryTask already was passing through {{epoch}} in {{solve()}}, which I think is the standard WCS codepath, so why epoch wasn't being applied already I don't fully understand.

laurenam:
`AstrometryTask`  was passing it to `loadPixelBox` which was NOT passing it along to `loadSkyCircle` which is where the correction happens...hence this ticket 😉

PR #144, opened 2021-06-09
DM-30490: Add magnitude outlier rejection to AstrometryTask.

erykoff:


python/lsst/meas/astrom/astrometry.py
```diff
@@ -358,16 +377,64 @@ def _matchAndFitWcs(self, refCat, sourceCat, goodSourceCat, refFluxField, bbox,
...
+        deltaMag = refMag - sourceMag
+        # Protect against negative fluxes and nans in the reference catalog.
+        goodDelta, = np.where(np.isfinite(deltaMag))
+        zp = np.median(deltaMag[goodDelta])
+        # Use median absolute deviation (MAD) for zpSigma
+        # Also require a minimum scatter to prevent floating-point errors from
+        # rejecting objects in zero-noise tests.
+        zpSigma = np.clip(1.4826*np.median(np.abs(deltaMag[goodDelta] - zp)), 1e-3, None)
+
+        self.log.debug(f"Rough zeropoint from astrometry matches is {zp:.4f} +/- {zpSigma:.4f}.")
```
laurenam, line 428: Just a fly-by comment...I'd actually like to see this as a `log.info` since it could be very useful to know in case things went horribly "wrong" with this zeropoint estimate (i.e. most likely something amiss with the source catalog itself).  We should normally have a "decent" estimate of what a photometric zeropoint should be, so having this logged here might help reveal seriously non-photometric visits (or something else funky going on).

python/lsst/meas/astrom/astrometry.py
```diff
@@ -417,20 +417,24 @@ def _removeMagnitudeOutliers(self, sourceFluxField, refFluxField, matchesIn):
...
         deltaMag = refMag - sourceMag
-        zp = np.median(deltaMag)
+        # Protect against negative fluxes and nans in the reference catalog.
+        goodDelta, = np.where(np.isfinite(deltaMag))
+        zp = np.median(deltaMag[goodDelta])
         # Use median absolute deviation (MAD) for zpSigma
         # Also require a minimum scatter to prevent floating-point errors from
         # rejecting objects in zero-noise tests.
-        zpSigma = np.clip(1.4826*np.median(np.abs(deltaMag - zp)), 1e-3, None)
+        zpSigma = np.clip(1.4826*np.median(np.abs(deltaMag[goodDelta] - zp)), 1e-3, None)
```
parejkoj, line 426: Can we replace this whole calculation with `scipy.stats.median_abs_deviation`, which has a `nan_policy` option?

python/lsst/meas/astrom/astrometry.py
parejkoj: I don't think that should be something we have to extract from the logs: that's a number that we should be recording as a metric (much like the astrometric scatter from the fit), so that we can readily access and plot it.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -57,6 +59,18 @@ class AstrometryConfig(RefMatchConfig):
...
+        dtype=bool,
+        doc=("If True then a rough zeropoint will be computed from matched sources "
+             "and outliers will be rejected in the iterations."),
+        default=False,
+    )
+    magnitudeOutlierRejectionNSigma = pexConfig.Field(
+        dtype=float,
+        doc=("Number of sigma (measured from the distribution) in magnitude "
+             "for a potential match to be rejected during iteration."),
+        default=3.0,
```
parejkoj, line 72: Instead of two config fields, could you just have the NSigma one and make it `optional` with `default=None`? That way, all the user has to do is set the value they want, and your `if` below just checks for whether that is `None`.

parejkoj, line 72: I think it would be good to be explicit in the docstring: "for a reference/source match to be rejected".

python/lsst/meas/astrom/astrometry.py
```diff
@@ -358,16 +377,64 @@ def _matchAndFitWcs(self, refCat, sourceCat, goodSourceCat, refFluxField, bbox,
...
+        # Use median absolute deviation (MAD) for zpSigma
+        # Also require a minimum scatter to prevent floating-point errors from
+        # rejecting objects in zero-noise tests.
+        zpSigma = np.clip(1.4826*np.median(np.abs(deltaMag[goodDelta] - zp)), 1e-3, None)
+
+        self.log.debug(f"Rough zeropoint from astrometry matches is {zp:.4f} +/- {zpSigma:.4f}.")
+
+        goodStars, = np.where(np.abs(deltaMag[goodDelta] - zp)
+                              <= self.config.magnitudeOutlierRejectionNSigma*zpSigma)
+        goodStars = goodDelta[goodStars]
```
parejkoj, line 432: With `nanmedian` and `median_abs_deviation` above, I think this becomes `goodStars = goodDelta[np.abs(deltaMag - zp) <= nSigma*zpSigma]`.

python/lsst/meas/astrom/astrometry.py
parejkoj: Please use the lazy-logging formatter for, so that the string doesn't have to be evaluated unless it is logged, e.g. `log.debug("Rough zeropoint from astrometry matches is %.4f +/- %.4f.", zp, zpSigma)`

python/lsst/meas/astrom/astrometry.py
```diff
@@ -358,16 +377,64 @@ def _matchAndFitWcs(self, refCat, sourceCat, goodSourceCat, refFluxField, bbox,
...
+        nOutlier = nMatch - goodStars.size
+        self.log.info(f"Removing {nOutlier} of {nMatch} magnitude outliers.")
```
parejkoj, line 434: How about: info("Removed %s magnitude outliers out of %s total astrometry matches.", nOutlier, nMatch)?

I think it would also be good to have a lazy-formatted debug-level message that includes some more information about those outliers (say, the min/max sigma of the removed outliers?)

python/lsst/meas/astrom/astrometry.py
```diff
@@ -358,16 +377,64 @@ def _matchAndFitWcs(self, refCat, sourceCat, goodSourceCat, refFluxField, bbox,
...
+
+        matchesOut = []
+        for matchInd in goodStars:
+            matchesOut.append(matchesIn[matchInd])
```
parejkoj, line 438: Though I am not responsible for it, I do want to apologize that our match lists require this sort of juggling.

tests/test_astrometryTask.py
```diff
@@ -162,6 +162,17 @@ def doTest(self, pixelsToTanPixels, order=3):
...
 
+        # try again, allowing magnitude outlier rejection.
+        config.doMagnitudeOutlierRejection = True
+        solverMagOutlierRejection = AstrometryTask(config=config, refObjLoader=self.refObjLoader)
+        self.exposure.setWcs(distortedWcs)
+        resultsMagOutlierRejection = solverMagOutlierRejection.run(
+            sourceCat=sourceCat,
+            exposure=self.exposure,
+        )
+        self.assertEqual(len(resultsMagOutlierRejection.matches), len(resultsRefSelect.matches))
```
parejkoj, line 173: I think all that this is testing is that we get the same number of matches whether we do outlier rejection or not? That doesn't seem like a useful test of the outlier rejection system itself. Could you set a tighter outlier threshold and check that some things actually got rejected? I don't know what this test data actually looks like.

There's a `maxAngSep` and `maxPixSep` calcluation above this that could be recycled (in a closure or free function, maybe?) to check the actual result.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -358,16 +377,64 @@ def _matchAndFitWcs(self, refCat, sourceCat, goodSourceCat, refFluxField, bbox,
...
             match_tolerance=matchRes.match_tolerance,
         )
+
+    def _removeMagnitudeOutliers(self, sourceFluxField, refFluxField, matchesIn):
+        """Remove magnitude outliers, computing a simple zeropoint.
+
+        Parameters
+        ----------
+        sourceFluxField : `str`
+            Field in source catalog for fluxes.
```
parejkoj, line 400: I think I want the docstring to be "for instrumental fluxes" here (certainly, that's what the class will be passing in).

parejkoj (changes requested):
I have some suggestions on logging and using numpy/scipy for nan-free calculations, and a comment on the tests.

I think you can squash the "nan protection" commit into the main one, since they both apply to the same method.

python/lsst/meas/astrom/astrometry.py
erykoff: I'm going to push back against this one.  Of all of the idiosyncrasies of the stack, I _love_ the `do*` pattern which makes it very clear which options are turned on and which are not.

tests/test_astrometryTask.py
erykoff: So the test catalogs are all 0-magnitude, perfect.  I didn't add any scatter to this, I just wanted to test here that my code ran smoothly and didn't reject anything.  Adding scatter to the test reference catalogs would be another whole thing that I was hoping to avoid.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -417,20 +417,24 @@ def _removeMagnitudeOutliers(self, sourceFluxField, refFluxField, matchesIn):
             refMag[i] = (match[0][refFluxField]*units.nJy).to_value(units.ABmag)
 
         deltaMag = refMag - sourceMag
-        zp = np.median(deltaMag)
+        # Protect against negative fluxes and nans in the reference catalog.
+        goodDelta, = np.where(np.isfinite(deltaMag))
```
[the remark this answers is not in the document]
erykoff: Unfortunately, I also need to protect against infinity as well as nan.  The astropy unit conversion will return `np.inf` when the flux is equal to 0.0.

python/lsst/meas/astrom/astrometry.py
erykoff: See above about protecting against nans being insufficient.  But I can use the scipy stats version instead of rolling my own.

tests/test_astrometryTask.py
erykoff: Upon further review, this is necessary to do all the proper checking, and I have added some outliers to the source and reference catalogs.

python/lsst/meas/astrom/astrometry.py
erykoff: Because of the indexing, this can't easily be done with a lazy computation, and I don't think it's necessary to go into some great computations just for a debug statement.

python/lsst/meas/astrom/astrometry.py
erykoff: With boolean arrays this doesn't work.  `goodDelta` has to be an index array, so I have left it as an index array but simplified the steps to make this a single line.

python/lsst/meas/astrom/astrometry.py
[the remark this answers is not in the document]
parejkoj: :cry:

python/lsst/meas/astrom/astrometry.py
```diff
@@ -358,16 +379,66 @@ def _matchAndFitWcs(self, refCat, sourceCat, goodSourceCat, refFluxField, bbox,
...
+            refMag[i] = (match[0][refFluxField]*units.nJy).to_value(units.ABmag)
+
+        deltaMag = refMag - sourceMag
+        # Protect against negative fluxes and nans in the reference catalog.
+        goodDelta, = np.where(np.isfinite(deltaMag))
+        zp = np.median(deltaMag[goodDelta])
+        # Use median absolute deviation (MAD) for zpSigma
+        # Also require a minimum scatter to prevent floating-point errors from
+        # rejecting objects in zero-noise tests.
+        zpSigma = np.clip(1.4826*scipy.stats.median_abs_deviation(deltaMag[goodDelta]),
```
parejkoj, line 428: I think you can avoid that constant by setting `scale="normal"` in the `median_abs_deviation` call.

python/lsst/meas/astrom/astrometry.py
parejkoj: What did you think about my reworded log message? As written, I feel like it might be a bit hard to get the meaning in the middle of a log.

tests/test_astrometryTask.py
```diff
@@ -162,6 +162,17 @@ def doTest(self, pixelsToTanPixels, order=3):
...
 
+        # try again, allowing magnitude outlier rejection.
+        config.doMagnitudeOutlierRejection = True
+        solverMagOutlierRejection = AstrometryTask(config=config, refObjLoader=self.refObjLoader)
+        self.exposure.setWcs(distortedWcs)
+        resultsMagOutlierRejection = solverMagOutlierRejection.run(
+            sourceCat=sourceCat,
+            exposure=self.exposure,
+        )
+        self.assertLess(len(resultsMagOutlierRejection.matches), len(resultsRefSelect.matches))
```
parejkoj, line 173: Don't you know exactly how many sources should be rejected, since you explicitly constructed the list? If you're not going to test the actual fit result, at least be more strict about testing the rejections.

parejkoj: Relatedly: I'd like the rejection test to check something close to the specific value of the nSigma config field, instead of `1000x`. There's some math going on in the new method, and it would be good to know that `nSigma` means what a user thinks it to mean.

tests/test_astrometryTask.py
```diff
@@ -211,6 +222,18 @@ def makeSourceCat(self, distortedWcs):
...
             src.set(sourceInstFluxKey, refObj.get(refFluxRKey))
             src.set(sourceInstFluxErrKey, refObj.get(refFluxRKey)/100)
+
+        # Make a few outlier fluxes for the outlier rejection tests.
+        # Otherwise, these fluxes are not used.
+        unResolved, = np.where(~refCat['resolved'])
+
+        sourceCat[sourceInstFluxKey][unResolved[0]] = 0.0
+        sourceCat[sourceInstFluxKey][unResolved[1]] = -1.0
+        sourceCat[sourceInstFluxKey][unResolved[2]] *= 1000.0
```
parejkoj, line 232: A comment would be good here, to note what each of these values should be testing. I think `0` and `-1` are to check the nan/inf rejection math, while 1000 is to check the outlier reject level?

python/lsst/meas/astrom/astrometry.py
parejkoj: Based on a brief survey on #dm-science-pipelines, it looks like several others prefer your approach, too.

tests/test_astrometryTask.py
```diff
@@ -162,6 +164,17 @@ def doTest(self, pixelsToTanPixels, order=3):
         )
         self.assertLess(len(resultsRefSelect.matches), len(results.matches))
 
+        # try again, allowing magnitude outlier rejection.
```
parejkoj, line 167: If you're going to keep this test--and given your new one, I don't know what point this serves--you should at least mention why you expect the same number of matches. If you could tell that setting the config actually changed the result, it might be worth keeping, but the test doesn't even tell us if the code was run (outside of looking at the coverage plot).

I think I'd advocate for restoring the code you added that made a few things outliers, and now you don't have to be careful about what kind of outlier, since all you care about is "did the outlier rejection actually run".

tests/test_astrometryTask.py
```diff
@@ -211,9 +224,59 @@ def makeSourceCat(self, distortedWcs):
...
+        refMag = np.full(nTest, 20.0)
+        srcMag = np.random.normal(size=nTest, loc=0.0, scale=1.0)
+
+        # Determine the sigma of the random sample
+        zp = np.median(refMag[: -4] - srcMag[: -4])
+        sigma = scipy.stats.median_abs_deviation(srcMag[: -4], scale='normal')
+
+        # Deliberately alter some magnitudes to be outliers.
+        srcMag[-3] = 4.1*sigma + (20.0 - zp)
+        srcMag[-4] = -4.1*sigma + (20.0 - zp)
```
parejkoj, line 261: Please make the connection between the `4.1` here, and the `magnitudeOutlierRejectionNSigma = 4.0` above explicit. Maybe by making this multiplier `magnitudeOutlierRejectionNSigma + 0.1`?

python/lsst/meas/astrom/astrometry.py
```diff
@@ -358,16 +379,68 @@ def _matchAndFitWcs(self, refCat, sourceCat, goodSourceCat, refFluxField, bbox,
...
+        refMag = np.zeros(nMatch)
+        for i, match in enumerate(matchesIn):
+            sourceMag[i] = -2.5*np.log10(match[1][sourceFluxField])
+            refMag[i] = (match[0][refFluxField]*units.nJy).to_value(units.ABmag)
+
+        deltaMag = refMag - sourceMag
+        # Protect against negative fluxes and nans in the reference catalog.
+        goodDelta, = np.where(np.isfinite(deltaMag))
+        zp = np.median(deltaMag[goodDelta])
+        # Use median absolute deviation (MAD) for zpSigma
```
parejkoj, line 425: period at end of sentence?

tests/test_astrometryTask.py
```diff
@@ -211,9 +224,59 @@ def makeSourceCat(self, distortedWcs):
             src.set(sourceCentroidKey, refObj.get(refCentroidKey))
             src.set(sourceInstFluxKey, refObj.get(refFluxRKey))
             src.set(sourceInstFluxErrKey, refObj.get(refFluxRKey)/100)
+
         return sourceCat
 
 
+class TestMagnitudeOutliers(lsst.utils.tests.TestCase):
+    def testMagnitudeOutlierRejection(self):
+        """Test rejection of magnitude outliers."""
```
parejkoj, line 233: If I'm reading the code correctly, all of these objects live at `(0,0)`, no? Please note in the docstring that this is only testing outlier rejection, not any other part of the matching or astrometry fitter.

tests/test_astrometryTask.py
```diff
@@ -211,9 +224,59 @@ def makeSourceCat(self, distortedWcs):
...
+        # Deliberately alter some magnitudes to be outliers.
+        srcMag[-3] = 4.1*sigma + (20.0 - zp)
+        srcMag[-4] = -4.1*sigma + (20.0 - zp)
+
+        refCat['refFlux'] = (refMag*units.ABmag).to_value(units.nJy)
+        srcCat['srcFlux'] = 10.0**(srcMag/(-2.5))
+
+        # Deliberately poison some reference fluxes.
+        refCat['refFlux'][-1] = np.inf
+        refCat['refFlux'][-2] = np.nan
```
parejkoj, line 268: Any particular reason you're putting these (and the srcMag changes above) at the end of the array instead of the beginning? Don't need to change it, I'm just curious.

parejkoj (approved):
Thank you for the new unittest: that's a lot better, and I like that it's a true unit test. A few more comments on it.

tests/test_astrometryTask.py
erykoff: I wanted to leave these tests as "null" tests to just show that the code does no harm if everything is perfect.  However, it's true that there's no other way to know that the code was actually run.  So I'll add in the bad guys.

tests/test_astrometryTask.py
erykoff: I have no idea, I just started by saying "I'll fit the first 96" and went down that road.

PR #145, opened 2021-06-16
DM-30701: Record initial astrometry matching statistics in exposure metadata.

erykoff:


python/lsst/meas/astrom/astrometry.py
```diff
@@ -287,6 +287,11 @@ def solve(self, exposure, sourceCat):
                 m.second.set(self.usedKey, True)
         exposure.setWcs(res.wcs)
 
+        # Record the scatter in the exposure metadata
+        md = exposure.getMetadata()
+        md['INIT_ASTROM_OFFSET_MEAN'] = tryMatchDist.distMean.asArcseconds()
+        md['INIT_ASTROM_OFFSET_STD'] = tryMatchDist.distStdDev.asArcseconds()
```
laurenam, line 293: As you anticipated(!), I do have a minor quibble with the variable names.  It's not at all clear to me what the "INIT" clause refers to (and my brain often goes to as-in "initial estimate heading into some iterative algorithm after which a final measurement will have been converged upon).  These are actually the FINAL versions of these metrics for SFM, and that's all the `calexp` is.  If, e.g., `jointcal` is later run, then this would be a more FINAL measurement...but we don't attach any external calibration info to the `calexp`s... So, I'd either just leave the `INIT_` off altogether (or maybe make it something like `SFM_`, but that may be confusing/taboo as an acronym, and spelling it out would be a bit much!)

erykoff: I like `SFM`.  I'll go with that, it's definitely better than `INIT`.

laurenam (approved):

PR #147, opened 2021-09-13
DM-30927: Set calexp WCS to None for failed astrometric fit

laurenam:


tests/test_astrometryTask.py
```diff
@@ -99,18 +98,30 @@ def testUsedFlag(self):
...
+        config.wcsFitter.maxScatterArcsec = 0.0  # To ensure a WCS failure
+        sourceSchema = afwTable.SourceTable.makeMinimalSchema()
+        measBase.SingleFrameMeasurementTask(schema=sourceSchema)  # expand the schema
+        # schema must be passed to the solver task constructor
+        solver = AstrometryTask(config=config, refObjLoader=self.refObjLoader, schema=sourceSchema)
+        sourceCat = self.makeSourceCat(self.tanWcs, sourceSchema=sourceSchema, doScatterCentroids=True)
+        results = solver.run(
+            sourceCat=sourceCat,
+            exposure=self.exposure,
+        )
```
erykoff, line 118: Is it also possible to test if we log a warning as expected?  (I actually don't know how to do that...)

python/lsst/meas/astrom/astrometry.py
```diff
@@ -251,71 +251,92 @@ def solve(self, exposure, sourceCat):
...
+            matches = res.matches
+            scatterOnSky = res.scatterOnSky
 
         # Record the scatter in the exposure metadata
         md = exposure.getMetadata()
-        md['SFM_ASTROM_OFFSET_MEAN'] = tryMatchDist.distMean.asArcseconds()
-        md['SFM_ASTROM_OFFSET_STD'] = tryMatchDist.distStdDev.asArcseconds()
+        if fitFailed:
+            md['SFM_ASTROM_OFFSET_MEAN'] = np.nan
+            md['SFM_ASTROM_OFFSET_STD'] = np.nan
```
erykoff, line 331: I'm actually uncertain about this one.  We do have a measurement of the OFFSET_MEAN and OFFSET_STD, it's just that they failed our threshold, right?  I think having these data recorded (how bad was it?) is valuable.

tests/test_astrometryTask.py
```diff
@@ -99,18 +98,30 @@ def testUsedFlag(self):
...
+        # Check that matches is set to None, the sourceCat coord cols are all
+        # set to nan and that the WCS attached to the exposure is set to None.
+        self.assertTrue(results.matches is None)
+        self.assertTrue(np.all(np.isnan(sourceCat["coord_ra"])))
+        self.assertTrue(np.all(np.isnan(sourceCat["coord_dec"])))
+        self.assertTrue(self.exposure.getWcs() is None)
```
erykoff, line 124: This should also be testing the metadata (though maybe not to check if they're nans, per my previous comment.)

tests/test_astrometryTask.py
timj: Yes. `with self.assertLogs()` (if it's a C++ lsst.log log message you have to use an additional bit of code to redirect the C++ logging to python logging)

erykoff (changes requested):
See comments on recording the stats from the bad fit.

tests/test_astrometryTask.py
laurenam: I don't think I can check the metadata you're looking for here.  I did add checks for:
```
self.assertTrue(results.scatterOnSky is None)
self.assertTrue(results.matches is None)
```
and I have a placeholder check for the metadata in `ci_hsc_gen3` as:
```
    def testMetadataForFailedAstrom(self):
        """Test that the metadata for a failed astrometic fit is set properly.

        Wait until Eli weighs in on what he wants this to be before committing.
        """
        calexpMetadata = self.butler.get("calexp.metadata", self.calexpMinimalDataId)
        self.assertTrue(np.isnan(calexpMetadata["SFM_ASTROM_OFFSET_MEAN"]))
        self.assertTrue(np.isnan(calexpMetadata["SFM_ASTROM_OFFSET_STD"]))
```

PR #149, opened 2021-10-10
DM-32129: Add a maximum mean offset threshold for SFM astrometry fit success

laurenam:


python/lsst/meas/astrom/astrometry.py
```diff
@@ -279,9 +288,14 @@ def solve(self, exposure, sourceCat):
...
-            "found %d matches with scatter = %0.3f +- %0.3f arcsec" %
+            "found %d matches with mean and scatter = %0.3f +- %0.3f arcsec" %
             (iterNum, len(tryRes.matches), tryMatchDist.distMean.asArcseconds(),
-                tryMatchDist.distStdDev.asArcseconds()))
+             tryMatchDist.distStdDev.asArcseconds()))
+        if tryMatchDist.distMean.asArcseconds() > self.config.maxMeanOffsetArcsec:
+            self.log.fatal("Assigning as a fit failure: mean offset on sky = %0.3f arcsec > %0.3f "
+                           "config.maxMeanOffsetArcsec" % (tryMatchDist.distMean.asArcseconds(),
+                                                           self.config.maxMeanOffsetArcsec))
+            raise RuntimeError
```
parejkoj, line 298: Is `RuntimeError` the the best choice for an error that will mark the fit as a failure? Should (do?) we have an LSST exception that means "this is an algorithmic fit failure, not a code failure"? That could be useful.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -279,9 +288,14 @@ def solve(self, exposure, sourceCat):
 
         self.log.info(
             "Matched and fit WCS in %d iterations; "
-            "found %d matches with scatter = %0.3f +- %0.3f arcsec" %
+            "found %d matches with mean and scatter = %0.3f +- %0.3f arcsec" %
             (iterNum, len(tryRes.matches), tryMatchDist.distMean.asArcseconds(),
-                tryMatchDist.distStdDev.asArcseconds()))
+             tryMatchDist.distStdDev.asArcseconds()))
+        if tryMatchDist.distMean.asArcseconds() > self.config.maxMeanOffsetArcsec:
+            self.log.fatal("Assigning as a fit failure: mean offset on sky = %0.3f arcsec > %0.3f "
```
parejkoj, line 295: Why not just "Fit failed: mean offset ..."? We're not assigning it as a failure, we're saying "it must be this good to be considered success."

python/lsst/meas/astrom/astrometry.py
```diff
@@ -279,9 +288,14 @@ def solve(self, exposure, sourceCat):
...
+                           "config.maxMeanOffsetArcsec" % (tryMatchDist.distMean.asArcseconds(),
```
parejkoj, line 296: "%" -> ","  - Use the lazy log formatter, instead of formatting the string directly.

tests/test_astrometryTask.py
```diff
@@ -112,6 +112,37 @@ def testUsedFlag(self):
...
+        sourceInstFluxKey = sourceSchema["slot_ApFlux_instFlux"].asKey()
+        sourceInstFluxErrKey = sourceSchema["slot_ApFlux_instFluxErr"].asKey()
+
+        for refObj in refCat:
+            src = sourceCat.addNew()
+            src.set(sourceCentroidKey, refObj.get(refCentroidKey))
+            src.set(sourceInstFluxKey, refObj.get(refFluxRKey))
+            src.set(sourceInstFluxErrKey, refObj.get(refFluxRKey)/100)
+
+        with self.assertRaises(RuntimeError):
```
parejkoj, line 143: Use `assertRaisesRegex` to ensure the contents of the exception message are what you want.

tests/test_astrometryTask.py
```diff
@@ -112,6 +112,37 @@ def testUsedFlag(self):
...
+        config.maxMeanOffsetArcsec = 0.0  # To ensure a "deemed" WCS failure
+        # schema must be passed to the solver task constructor
+        solver = AstrometryTask(config=config, refObjLoader=self.refObjLoader, schema=sourceSchema)
+        sourceCat = afwTable.SourceCatalog(sourceSchema)
+        sourceCat.reserve(len(refCat))
+        sourceCentroidKey = afwTable.Point2DKey(sourceSchema["slot_Centroid"])
+        sourceInstFluxKey = sourceSchema["slot_ApFlux_instFlux"].asKey()
+        sourceInstFluxErrKey = sourceSchema["slot_ApFlux_instFluxErr"].asKey()
+
+        for refObj in refCat:
```
parejkoj, line 137: Why not just use the existing `makeSourceCat` method in this test? I think that could replace a bunch of your catalog management in this test.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -60,6 +60,15 @@ class AstrometryConfig(RefMatchConfig):
         default=0.001,
         min=0,
     )
+    maxMeanOffsetArcsec = pexConfig.RangeField(
+        doc="Maximum mean offset (in arcsec) of a WCS fit beyond which the fit is considered "
+            "a failure.  Default is set to maximum tolerated by the external global calibration "
+            "(i.e. jointcal) step for conceivable recovery.  Appropriate value will be specific "
```
parejkoj, line 66: I think "specific" is unneeded here?

tests/test_astrometryTask.py
laurenam: Thanks for pointing that out!  I guess I very lazily just copy-pasta'd from the `testUsedFlag` test above and didn't pay attention to that method.  I've actually updated it for ease of reuse and to add some optional extra functionality.  I updated both my new test and `testUsedFlag`  above to use that function.

python/lsst/meas/astrom/astrometry.py
```diff
@@ -60,6 +60,15 @@ class AstrometryConfig(RefMatchConfig):
         default=0.001,
         min=0,
     )
+    maxMeanDistanceArcsec = pexConfig.RangeField(
+        doc="Maximum mean on-sky distance (in arcsec) of a WCS fit beyond which the fit is "
+            "considered a failure.  Default is set to maximum tolerated by the external global "
```
erykoff, line 65: I think this should say what exception is raised if it fails (`TaskError` I guess).

python/lsst/meas/astrom/astrometry.py
```diff
@@ -279,9 +288,14 @@ def solve(self, exposure, sourceCat):
...
             "Matched and fit WCS in %d iterations; "
-            "found %d matches with scatter = %0.3f +- %0.3f arcsec" %
-            (iterNum, len(tryRes.matches), tryMatchDist.distMean.asArcseconds(),
-                tryMatchDist.distStdDev.asArcseconds()))
+            "found %d matches with on-sky distance mean and scatter = %0.3f +- %0.3f arcsec",
+            iterNum, len(tryRes.matches), tryMatchDist.distMean.asArcseconds(),
+            tryMatchDist.distStdDev.asArcseconds())
+        if tryMatchDist.distMean.asArcseconds() > self.config.maxMeanDistanceArcsec:
+            raise pipeBase.TaskError(
+                "Assigning as an astrometry failure: mean on-sky distance = %0.3f arcsec > %0.3f "
```
erykoff, line 296: Maybe rephrase to something like "Fatal astrometry failure detected: "

tests/test_astrometryTask.py
```diff
@@ -112,6 +99,19 @@ def testUsedFlag(self):
...
+        """If the astrometric fit does not satisfy the maxMeanDistanceArcsec
+        threshold, ensure task raises an lsst.pipe.base.TaskError.
+        """
+        self.exposure.setWcs(self.tanWcs)
+        config = AstrometryTask.ConfigClass()
+        config.maxMeanDistanceArcsec = 0.0  # To ensure a "deemed" WCS failure
+        solver = AstrometryTask(config=config, refObjLoader=self.refObjLoader)
+        sourceCat = self.makeSourceCat(self.tanWcs, doScatterCentroids=True)
+
+        with self.assertRaisesRegex(pipeBase.TaskError, "Assigning as an astrometry failure"):
```
erykoff, line 112: ...and if you rephrase, please update this as well!

tests/test_astrometryTask.py
```diff
@@ -203,25 +203,43 @@ def doTest(self, pixelsToTanPixels, order=3):
...
         sourceInstFluxKey = sourceSchema["slot_ApFlux_instFlux"].asKey()
         sourceInstFluxErrKey = sourceSchema["slot_ApFlux_instFluxErr"].asKey()
 
+        scatterFactor = 1.0
+        np.random.seed(12345)
+
         sourceCat.reserve(len(refCat))
         for refObj in refCat:
+            if doScatterCentroids:  # add some small random offsets to source centroids
+                scatterFactor = np.random.uniform(0.999, 1.001)
```
erykoff, line 238: This is totally irrelevant to the ticket, but seeing this loop over sources instead of doing arrays is making me a bit 🤢.  No need to change, but now I'm on record.

python/lsst/meas/astrom/astrometry.py
erykoff: Reading this again, I think that "distance of a WCS fit" doesn't quite make sense.  Perhaps something like "Maximum mean on-sky distance (in arcsec) between the sources and matched reference stars post-fit.  Mean distance greater than this threshold raises a TaskError and the WCS fit is considered a failure.  Default is set ..."

tests/test_astrometryTask.py
laurenam: Well, if only your gag-reflex kicked in when you were last updating this file 😜. I gave it my best effort (I got a bit hung up on setting the centroids...), so now you get to review my changes to make sure I didn't mess it up (the tests do pass, and I did check that mine failed in the same manner with the un-updated `with self.assertRaisesRegex`, so hopefully all is ok)!

PR #150, opened 2021-10-12
DM-31801: Use lsst prefix for logger and use python logging

timj:


python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py
```diff
@@ -204,8 +204,8 @@ def match(self, source_array, n_check, n_match, n_agree,
             max_dist_rad=None,)
 
         if n_source <= 0:
-            self.log.warn("Source object array is empty. Unable to match. "
-                          "Exiting matcher.")
+            self.log.warning("Source object array is empty. Unable to match. "
+                             "Exiting matcher.")
```
parejkoj, line 208: Can this all go on one line?

parejkoj (approved):
Single suggestion, otherwise this looks good.

PR #152, opened 2021-10-22
DM-32008: Wrap `test_spoke` in C++/pybind11

morriscb:


python/lsst/meas/astrom/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,128 @@
...
+
+namespace py = pybind11;
+
+namespace lsst {
+namespace meas {
+namespace astrom {
+
+namespace {
+
+int testSpoke(double cos_theta_src, double sin_theta_src, py::array_t<double> ref_ctr, int ref_ctr_id,
```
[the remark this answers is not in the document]
morriscb: I can change it to check_spoke

morriscb:
Finished things up mostly except for one compiler warning.

parejkoj:
Please squash your commits.

include/lsst/meas/astrom/pessimisticPatternMatcherUtils.h
```diff
@@ -0,0 +1,69 @@
...
+ *     that our center id object is paired with.
+ * @param[in] reference_array Array of three vectors representing the locations
+ *     of all reference objects.
+ * @param[in] src_sin_tol Sine of tolerance allowed between source and
+ *     reference spoke opening angles.
+ * @return ID of the candidate reference object successfully matched or -1 if
+ *     no match is found.
+ */
+int check_spoke(double cos_theta_src, double sin_theta_src, ndarray::Array<double, 1, 1> const& ref_ctr,
+                uint16_t ref_ctr_id, ndarray::Array<double, 1, 1> const& proj_ref_ctr_delta,
```
parejkoj, line 62: Is `ref_ctr_id` really a uint16? That means the refcat can only have 65k entries. I'd expect this and `ref_id_array` to both be `std::size_t`, since they are array indexes.

python/lsst/meas/astrom/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,46 @@
...
+
+namespace py = pybind11;
+
+namespace lsst {
+namespace meas {
+namespace astrom {
+
+PYBIND11_MODULE(pessimisticPatternMatcherUtils, mod) {
+    mod.def("check_spoke", &check_spoke,
+            "Test the opening angle between the spokes of our source pattern against the reference.");
```
parejkoj, line 41: Huh, I wasn't aware that you could add a python docstring like this. We don't mention it in the dev guide, but we probably should!

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,115 @@
...
+        /// the source angle. Assumes that the delta between the two is
+        /// small.
+        if (cos_sq_comparison > src_sin_tol * src_sin_tol) {
+            continue;
+        }
+        /// The cosine tests the magnitude of the angle but not
+        /// its direction. To do that we need to know the sine as well.
+        /// This cross product calculation does that.
+        double cross_ref[3];
+        cross_ref[0] =
```
parejkoj, line 84: You should be able to do all this with Eigen, but I'm not positive about the exact syntax. Something like:

```
Array<double, 1, 2> cross_ref = proj_ref_delta.asEigenMatrix().cross(proj_ref_ctr_delta.asEigenMatrix()/geom_dist_ref);
```

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,115 @@
...
+                (proj_ref_delta[1] * proj_ref_ctr_delta[2] - proj_ref_delta[2] * proj_ref_ctr_delta[1]) /
+                geom_dist_ref;
+        cross_ref[1] =
+                (proj_ref_delta[2] * proj_ref_ctr_delta[0] - proj_ref_delta[0] * proj_ref_ctr_delta[2]) /
+                geom_dist_ref;
+        cross_ref[2] =
+                (proj_ref_delta[0] * proj_ref_ctr_delta[1] - proj_ref_delta[1] * proj_ref_ctr_delta[0]) /
+                geom_dist_ref;
+        double sin_theta_ref =
+                cross_ref[0] * ref_ctr[0] + cross_ref[1] * ref_ctr[1] + cross_ref[2] * ref_ctr[2];
```
parejkoj, line 94: And again, with Eign, something like:

```
double sin_theta_ref = cross_ref.asEigenMatrix().dot(ref_ctr.asEigenMatrix());
```

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,115 @@
...
+            sin_comparison = (sin_theta_src - sin_theta_ref) / src_sin_tol;
+        } else {
+            sin_comparison = (sin_theta_src - sin_theta_ref) / cos_theta_ref;
+        }
+
+        if (abs(sin_comparison) > src_sin_tol) {
+            continue;
+        }
+        /// Return the correct id of the candidate we found.
+        return ref_id;
```
parejkoj, line 108: I feel like this and the above `if` should be reversed; "if the thing is found, return it, otherwise next iteration", thus no `continue` needed.

parejkoj (changes requested):
A number of array cleanups: Eigen and ndarray can do a bunch of things to clean this up.

Just to check: you ran clang-format on this?

python/lsst/meas/astrom/pessimisticPatternMatcherUtils.cc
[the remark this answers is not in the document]
morriscb: Changed it to check_spoke

include/lsst/meas/astrom/pessimisticPatternMatcherUtils.h
morriscb: It is. I'll change the other two types and see if it compiles.

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,115 @@
...
+namespace meas {
+namespace astrom {
+
+int check_spoke(double cos_theta_src, double sin_theta_src, ndarray::Array<double, 1, 1> const& ref_ctr,
+                uint16_t ref_ctr_id, ndarray::Array<double, 1, 1> const& proj_ref_ctr_delta,
+                double proj_ref_ctr_dist_sq, ndarray::Array<long int, 1, 1> const& ref_dist_idx_array,
+                ndarray::Array<uint16_t, 1, 1> const& ref_id_array,
+                ndarray::Array<double, 2, 1> const& reference_array, double src_sin_tol) {
+    /// Loop over our candidate reference objects.
+    for (auto idx = ref_dist_idx_array.begin(); idx != ref_dist_idx_array.end(); idx++) {
```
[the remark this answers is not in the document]
morriscb: How does this work exactly? `idx` here controls how we move through ref_id_array. The `idx` isn't strictly increasing.

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,115 @@
...
+        /// Compute the delta vector from the pattern center.
+        unsigned int ref_id = ref_id_array[*idx];
+        double ref_delta[3];
```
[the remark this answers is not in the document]
morriscb: Will do. However I will point out that the last number of is the number of contiguous dimensions. So you can't have 1, 2. This would be 1, 1.

include/lsst/meas/astrom/pessimisticPatternMatcherUtils.h
```diff
@@ -0,0 +1,69 @@
...
+ * @param[in] ref_id_array Array of id lookups into the master reference array
+ *     that our center id object is paired with.
+ * @param[in] reference_array Array of three vectors representing the locations
+ *     of all reference objects.
+ * @param[in] src_sin_tol Sine of tolerance allowed between source and
+ *     reference spoke opening angles.
+ * @return ID of the candidate reference object successfully matched or -1 if
+ *     no match is found.
+ */
+int check_spoke(double cos_theta_src, double sin_theta_src, ndarray::Array<double, 1, 1> const& ref_ctr,
```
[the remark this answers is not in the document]
morriscb: I've tested this while attempting to run and this is the contiguous settings that work.

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,115 @@
...
+        double proj_ref_delta[3];
+        ref_delta[0] = reference_array[ref_id][0] - ref_ctr[0];
+        ref_delta[1] = reference_array[ref_id][1] - ref_ctr[1];
+        ref_delta[2] = reference_array[ref_id][2] - ref_ctr[2];
+
+        double ref_dot = ref_delta[0] * ref_ctr[0] + ref_delta[1] * ref_ctr[1] + ref_delta[2] * ref_ctr[2];
+        proj_ref_delta[0] = ref_delta[0] - ref_dot * ref_ctr[0];
```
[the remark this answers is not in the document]
morriscb: Looks good. Again, it can't be 1, 2 as you can't have a second contiguous dimension.

src/pessimisticPatternMatcherUtils.cc
morriscb: I'll give these a shot on one or two of them and see if I can get it to compile/run. If I can't though, I'm not too uncomfortable with just leaving it for now.

morriscb:
Clang format is running every time I save the file, yes.

src/pessimisticPatternMatcherUtils.cc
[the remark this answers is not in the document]
parejkoj: What do you mean "isn't strictly increasing?" You do `idx++` and then do nothing else with it other than dereference the array, so a range-based for loop is just what you want.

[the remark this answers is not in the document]
morriscb: If you look again, I'm de-referencing a different array.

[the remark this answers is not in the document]
parejkoj: Ohhhhhh! Ok, that's probably worth a descriptive comment, since how the indexes of those two arrays relate is not obvious.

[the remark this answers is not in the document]
morriscb: Here's the the code that produces the index array. https://github.com/lsst/meas_astrom/blob/903eb1c80716f581a39b5935b9d6f8443b876770/python/lsst/meas/astrom/pessimistic_pattern_matcher_b_3D.py#L574

It is a view into the id array ordered by increasing difference between the source distance and reference distance array.

src/pessimisticPatternMatcherUtils.cc
morriscb: So, I attempted something similar to this, however, it seems cross is only available for Eigen::Vector3d objects so I would have to double convert as there doesn't seem to be a direct ndarray->Eigen vector method. Options are: leave this as is or double convert the ndarrays from Matrix to Vector3d and store in the method for later use. What do ya think?

src/pessimisticPatternMatcherUtils.cc
morriscb: Done.

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,105 @@
...
+namespace meas {
+namespace astrom {
+
+int check_spoke(double cos_theta_src, double sin_theta_src, ndarray::Array<double, 1, 1> const& ref_ctr,
+                uint16_t ref_ctr_id, ndarray::Array<double, 1, 1> const& proj_ref_ctr_delta,
+                double proj_ref_ctr_dist_sq, ndarray::Array<long int, 1, 1> const& ref_dist_idx_array,
+                ndarray::Array<uint16_t, 1, 1> const& ref_id_array,
+                ndarray::Array<double, 2, 1> const& reference_array, double src_sin_tol) {
+    /// Loop over our candidate reference objects. ref_dist_idx_array is a view into ref_id_array
+    /// and is not the same length as ref_id_array.
```
parejkoj, line 41: All of the comments should be double slash, not triple.

morriscb: Swear I saw tripled asked for somewhere in the docs but can't find it now. Will change it back.

src/pessimisticPatternMatcherUtils.cc
morriscb: Do you have a code example for this? Previously I had to dig through ndarray docs to find out to use asEigenMatrix properly.

parejkoj: https://ndarray.github.io/ndarray/eigen_8h.html#ad1573ab63904bc25bde732c1f6161014

morriscb: Done. Worked with @TallJimbo on getting the cross product to work. The cross function has very specific needs in order to be used.

src/pessimisticPatternMatcherUtils.cc
```diff
@@ -0,0 +1,101 @@
...
+#include <Eigen/Dense>
+#include "ndarray/eigen.h"
+#include "lsst/meas/astrom/pessimisticPatternMatcherUtils.h"
+
+namespace lsst {
+namespace meas {
+namespace astrom {
+
+int check_spoke(double cos_theta_src, double sin_theta_src, ndarray::Array<double, 1, 1> const& ref_ctr,
+                uint16_t ref_ctr_id, ndarray::Array<double, 1, 1> const& proj_ref_ctr_delta,
```
parejkoj, line 36: Hang on, I don't see ref_ctr_id used in the C++ code below at all?

include/lsst/meas/astrom/pessimisticPatternMatcherUtils.h
```diff
@@ -0,0 +1,69 @@
...
+ * @param[in] ref_ctr_id id lookup of the ref_ctr into the master reference
+ *     array.
+ * @param[in] proj_ref_ctr_delta Plane projected first spoke in the reference
+ *     pattern using the pattern center as normal.
+ * @param[in] proj_ref_ctr_dist_sq Squared length of the projected vector.
+ * @param[in] ref_dist_idx_array Indices sorted by the delta distance between
+ *     the source spoke we are trying to test and the candidate reference
+ *     spokes.
+ * @param[in] ref_id_array Array of id lookups into the master reference array
+ *     that our center id object is paired with.
```
parejkoj, line 53: Please add a note here about why this is uint16 (as you said, the higher level code trims the arrays due to memory limits): it's generally expected that indexes are size_t.

parejkoj (approved):
Remove ref_ctr_id, add a note about why ref_id_array is uint16, and you're good to go.

morriscb:
Done! Take a look at the comment for yourself if you want.

PR #153, opened 2021-10-25
DM-32300: Simplify `_build_distances_and_angles` to only use numpy arrays

morriscb:


kherner (approved):
Looks fine; any particular reason for float32 and uint16 as opposed to float64 and uint32?

morriscb:
It's for memory. The distance array doesn't need to be more precise than 32. The other array is 16 is the limit the size of the these pre-computed arrays and make the ref_id a bit smaller in memory.

PR #157, opened 2021-12-01
DM-32034: Create MatchProbabilisticTask

taranu:


python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -0,0 +1,437 @@
...
+    """
+    config: MatchProbabilisticConfig
+
+    def __init__(
+            self,
+            config: MatchProbabilisticConfig,
+    ):
+        self.config = config
+
+    def match(
```
morriscb, line 272: Code could use some inline comments describing what is happening at a give step.

python/lsst/meas/astrom/match_probabilistic_task.py
```diff
@@ -0,0 +1,174 @@
...
+import pandas as pd
+from typing import Dict, Set, Tuple
+
+__all__ = ['MatchProbabilisticTask']
+
+
+class MatchProbabilisticTask(pipeBase.Task):
+    """Run MatchProbabilistic on a reference and target catalog covering the same tract.
+    """
+    ConfigClass = MatchProbabilisticConfig
```
morriscb, line 39: Are there no configs specific to the task? Would it make more sense to have the prob matcher be a configurable subtask of the main task? It's fine if not.

python/lsst/meas/astrom/matcher_probabilistic.py
morriscb: Also, would it make sense to break some of this function up into sub sections? That would certainly make the testing easier.

morriscb (changes requested):
I think my main comment is to break up the parts of the match method in matcher_probabilistic into some smaller methods sub methods. As add some comments to explain the more complicated parts.

python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -0,0 +1,437 @@
...
+            n_found = len(found)
+            if n_found > 0:
+                chi = (
+                    (data_target.iloc[found].values - data_ref.iloc[index_n].values)
+                    / errors_target.iloc[found].values
+                )
+                finite = np.isfinite(chi)
+                n_finite = np.sum(finite, axis=1)
+                chisq_good = n_finite >= config.match_n_finite_min
+                if np.any(chisq_good):
```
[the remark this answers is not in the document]
taranu: Hmm, I suppose that's true. I guess I could add a min/max/default error or something, but then the user can also just modify the catalogs before passing them in.

python/lsst/meas/astrom/matcher_probabilistic.py
taranu: I don't think there would be much benefit in breaking it up, as any of the sub-functions I could think of defining would have large numbers of arguments/return values but still not really make sense to call in any other context.

python/lsst/meas/astrom/match_probabilistic_task.py
taranu: Just to clarify, I could have folded the Matcher into the Task instead of having a separate class, but this way the Matcher has no stack dependencies besides pex_config (which I believe can be installed as a standalone package itself), and in principle can be imported independently without building the package. Whether that ever ends up being useful to anyone remains to be seen.

python/lsst/meas/astrom/matcher_probabilistic.py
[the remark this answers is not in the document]
morriscb: Would it be possible to only compare the non-NaN columns?

python/lsst/meas/astrom/match_probabilistic_task.py
morriscb: Sounds good. Like I said on the other repo, I think you already take care of this by having the configurable in the pipeline class.

morriscb (approved):

python/lsst/meas/astrom/matcher_probabilistic.py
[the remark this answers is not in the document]
taranu: It's already based on the non-NaN columns, and will succeed as long as there are more than `n_finite` finite chi values. It's just that it doesn't distinguish between measurements vs errors being NaN (but again, the user can modify the input catalogs if they care, and we really shouldn't be producing finite values with NaN errors in the first place).

python/lsst/meas/astrom/match_probabilistic_task.py
```diff
@@ -0,0 +1,174 @@
...
+                             + len(config.columns_target_select_false)) > 0
+        if select_additional:
+            if select_target is None:
+                select_target = np.ones(len(catalog_target), dtype=bool)
+            for column in config.columns_target_select_true:
+                select_target &= catalog_target[column].values
+            for column in config.columns_target_select_false:
+                select_target &= ~catalog_target[column].values
+
+        logger.info(f'Beginning MatcherProbabilistic.match with {np.sum(select_ref)}/{len(select_ref)}'
```
timj, line 126: Please do not use f-strings with loggers. The logger must use `%` formatting and must be given as parameters at the end. The reason is to defer the stringification until we know we need it. There are four other places (at least) that need to be fixed.

PR #169, opened 2022-08-15
DM-35904: More gen2 cleanups

timj:


python/lsst/meas/astrom/directMatch.py
```diff
@@ -31,8 +33,8 @@ class DirectMatchTask(Task):
 
     Parameters
     ----------
-    butler : `lsst.daf.persistence.Butler`
-        Data butler containing the relevant reference catalog data.
+    butler : `None`
+        Compatibility parameter. Should not be used.
     refObjLoader : `lsst.meas.algorithms.LoadReferenceObjectsTask` or `None`
```
erykoff, line 38: This should be `lsst.meas.algorithms.ReferenceObjectLoader`, and below in `setRefObjLoader` that should only take `lsst.meas.algorithms.ReferenceObjectLoader` and the docstring can be simplified.

erykoff (approved):
Also in this file, `setRefObjLoader` needs a docstring cleanup; it should only take a `lsst.meas.algorithms.ReferenceObjectLoader` instance.

python/lsst/meas/astrom/directMatch.py
timj: I think I've cleaned that up now.

PR #170, opened 2022-08-31
DM-28329: remove joinMatchList test and cleanup to use exposure info properties

parejkoj:


timj (approved):
Looks like a nice clean up. Can you please fix the confusing PR title -- the changes here have nothing to do with jyear or EPOCH. It was initially confusing.

PR #174, opened 2023-01-30
Tickets/DM-32197: Cleanup lsstDebug references in Task docs

bsmartradio:
lsstDebug references are now linked to the lsstDebug docs rather than indirectly, as well as some minor readability edits.

Removed extra examples which were outdated in the documents and not accurate.

PR #175, opened 2023-02-03
DM-28969: Cleanup lsstDebug references in Task docs 

bsmartradio:
lsstDebug is now reference-able, so links have been added to the documents. A number of examples in the documentation were no longer valid or contained incorrect information and were removed.


bsmartradio:
The original PR review was done on an incorrectly named ticket, DM-32197. The original PR review can be seen [here](https://github.com/lsst/meas_astrom/pull/174).

PR #176, opened 2023-02-11
DM-37804: Use pybind wrappers, merge into one pybind shared lib

mwittgen:


python/lsst/meas/astrom/SConscript
```diff
@@ -1,10 +1,15 @@
...
-scripts.BasicSConscript.pybind11([
-    "makeMatchStatistics",
-    "matchOptimisticB/matchOptimisticB",
-    "polynomialTransform",
-    "scaledPolynomialTransformFitter",
-    "sipTransform",
-    "pessimisticPatternMatcherUtils",
-], addUnderscore=False)
+scripts.BasicSConscript.python(['_meas_astromLib'], [
+    '_meas_astromLib.cc',
```
erykoff, line 4: Not that it really matters (it's an underscore non-public interface), but I think this should be named `_measAstromLib` which is parallel to `meas_base` and `meas_algorithms`, etc.

python/lsst/meas/astrom/sip/__init__.py
```diff
@@ -21,11 +21,7 @@
 #
 
 import lsst.afw.geom  # for Box2I
-
-from .leastSqFitter1d import *
-from .leastSqFitter2d import *
-from .createWcsWithSip import *
-from .matchSrcToCatalogue import *
+from .._meas_astromLib import *
```
erykoff, line 24: I don't think the `sip` namespace should lift all of the library functions, but just the sip ones.

erykoff (changes requested):
I think the sip import needs to be fixed, and perhaps the library name should change (but it's internal).
One thing that I was a little less certain about is putting _everything_ in the same library, if they were separate before.  But since the base import before would import everything including the `sip` and `matchOptimisticB` code, I think this is fine.

erykoff (approved):

PR #177, opened 2023-03-01
DM-37683: Updated _DefaultName to match conventions

erinleighh:


parejkoj (approved):
Thanks. Trivial change, but we want to be consistent on these.

PR #181, opened 2023-06-23
DM-38546: fixes while implementing new CalibrateImageTask

parejkoj:


python/lsst/meas/astrom/ref_match.py
```diff
@@ -50,9 +50,16 @@ class RefMatchConfig(pexConfig.Config):
...
+    doSourceSelection = pexConfig.Field(
+        dtype=bool,
+        doc="Select sources to be matched with `sourceSelector`?"
+            " Set to False if you want to use exactly the sources that are passed in.",
+        default=True
+    )
     sourceSelector = sourceSelectorRegistry.makeField(
         doc="How to select sources for cross-matching.",
         default="science",
+        optional=True
```
TallJimbo, line 62: I'm normally in favor of adding `doX` options since there's such a strong precedent for it, but in this case setting the registry field selection to `None` seems to express the desired behavior plenty well enough, and unless we add something to `validate` to check `doSourceSelection is False` if and only if `sourceSelector is None`, we're messing up configuration comparisons a bit (though surely not for the first time), by making it so two functionally-equivalent configs with different `sourceSelector` values but `doSourceSelection=False` compare as unequal.

It's a pity the `DirectMatchConfigWithoutLoader.sourceSelector` field is a `ConfigurableField` and can't be optional, though, and consistency is also desirable.

What do you think about having a do-nothing `SourceSelector` implementation instead of any of these config field changes?

python/lsst/meas/astrom/astrometry.py
```diff
@@ -118,8 +119,8 @@ class AstrometryTask(RefMatchTask):
     ConfigClass = AstrometryConfig
     _DefaultName = "astrometricSolver"
 
-    def __init__(self, refObjLoader, schema=None, **kwargs):
-        RefMatchTask.__init__(self, refObjLoader, **kwargs)
+    def __init__(self, refObjLoader=None, schema=None, **kwargs):
+        RefMatchTask.__init__(self, refObjLoader=refObjLoader, **kwargs)
```
TallJimbo, line 123: It'd be nice to actually deprecate this argument, but certainly out of scope for this ticket.

TallJimbo (approved):

python/lsst/meas/astrom/astrometry.py
parejkoj: I'm not sure we actually want to deprecate it: I could use uses of it in e.g. notebooks.

python/lsst/meas/astrom/ref_match.py
parejkoj: I've added a `NullSourceSelectorTask` to meas_algorithms and reverted this change.

Conceptually, I don't like it because you should be able to truly do *nothing*, not even create a numpy array of bools, but that's probably a small enough thing to not be a concern in practice. But yes, I do wish we could just set those registry fields to None and be done with it; revising our uses of them to all use optional registries would be a nice deprecation step, but might be hard to do in practice.

PR #183, opened 2023-07-13
tickets/DM-39836: Change the algorithm FitAffineWcsTask uses

natelust:
This commit changes FitAffineWcsTask to directly solve for affine and shift parameters by solving a matrix equation instead of creating wcs objects while iterating in a solver. This also absorbs linear shifts inside the linear transforms instead of solving for a sphere point offset in the initial wcs.

### DIFF
--- a/python/lsst/meas/astrom/fitAffineWcs.py
+++ b/python/lsst/meas/astrom/fitAffineWcs.py
@@ -24,64 +24,17 @@ __all__ = ["FitAffineWcsTask", "FitAffineWcsConfig", "TransformedSkyWcsMaker"]
 
 import astshim
 import numpy as np
-from scipy.optimize import least_squares
+from scipy.linalg import lstsq
 
 from lsst.afw.geom import makeSkyWcs, SkyWcs
 import lsst.afw.math
-from lsst.geom import Point2D, degrees, arcseconds, radians
+from lsst.geom import Point2D, radians
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 from lsst.utils.timer import timeMethod
 
 from ._measAstromLib import makeMatchStatisticsInRadians
 from .setMatchDistance import setMatchDistance
-
-
-def _chiFunc(x, refPoints, srcPixels, wcsMaker):
-    """Function to minimize to fit the shift and rotation in the WCS.
-
-    Parameters
-    ----------
-    x : `numpy.ndarray`
-        Current fit values to test. Float values in array are:
-
-        - ``bearingTo``: Direction to move the wcs coord in.
-        - ``separation``: Distance along sphere to move wcs coord in.
-        - ``affine0,0``: [0, 0] value of the 2x2 affine transform matrix.
-        - ``affine0,1``: [0, 1] value of the 2x2 affine transform matrix.
-        - ``affine1,0``: [1, 0] value of the 2x2 affine transform matrix.
-        - ``affine1,1``: [1, 1] value of the 2x2 affine transform matrix.
-    refPoints : `list` of `lsst.afw.geom.SpherePoint`
-        Reference object on Sky locations.
-    srcPixels : `list` of `lsst.geom.Point2D`
-        Source object positions on the pixels.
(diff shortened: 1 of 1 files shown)


python/lsst/meas/astrom/fitAffineWcs.py
```diff
@@ -163,52 +116,41 @@ def fitWcs(self,
...
-        # direction and separation between the reference and the sources.
-        for match in matches:
+        # Grab the initial transformations going back from sky coordinates
+        # and forward from pixel coordinates.
+        back = wcsMaker.frameDict.getMapping(wcsMaker.frameMax, wcsMaker.frameMax-1)
+        forward = wcsMaker.lastMapBeforeSky
+
+        # Create containers for the data going into the Affine fit (i.e.
+        # solving Ax=b where x will be the affine parameters and a linear
+        # shift.
```
TallJimbo, line 126: Missing close-`)`.

This would also be a good place to document how you've ordered the affine transform parameters when flattening the matrix and offset into a 6-element vector.

And to be super precise, instead of just saying "solving Ax=b" you might want to say "minimizing ||Ax - b||" or otherwise referencing least-squares.

python/lsst/meas/astrom/fitAffineWcs.py
```diff
@@ -163,52 +116,41 @@ def fitWcs(self,
...
+        A = np.zeros((len(matches)*2, 6), dtype=float)
+        b = np.empty(len(matches)*2, dtype=float)
+
+        # Constant terms related to the shit in and and y parameters
```
TallJimbo, line 130:
```suggestion
        # Constant terms related to the shift in x and y parameters
```

python/lsst/meas/astrom/fitAffineWcs.py
```diff
@@ -163,52 +116,41 @@ def fitWcs(self,
...
+        A[:len(matches)*2:2, 4] = 1
+        A[1:len(matches)*2+1:2, 5] = 1
```
TallJimbo, line 132: I think
```
A[::2, 4] = 1
A[1::2, 5] = 1
```
is equivalent to the above, and more readable.

python/lsst/meas/astrom/fitAffineWcs.py
```diff
@@ -163,52 +116,41 @@ def fitWcs(self,
...
-            x0=[offsetDir, offsetDist, 1., 1e-8, 1e-8, 1.],
-            args=(refPoints, srcPixels, wcsMaker),
-            method='dogbox',
-            bounds=[[-360, -np.inf, -np.inf, -np.inf, -np.inf, -np.inf],
-                    [360, np.inf, np.inf, np.inf, np.inf, np.inf]],
-            ftol=2.3e-16,
-            gtol=2.31e-16,
-            xtol=2.3e-16)
-        self.log.debug("Best fit: Direction: %.3f, Dist: %.3f, "
+            val = forward.applyForward(srcCentroid)
```
TallJimbo, line 140: I'm pretty sure we could get another speedup by vectorizing these calls to `applyForward` if the profile says that'd help. The usefulness of that is limited by the fact that we're starting with a `ReferenceMatchVector` rather than something with array-column coordinates, though.

TallJimbo (approved):

python/lsst/meas/astrom/fitAffineWcs.py
natelust: I agree. Those were left from previous slicing operations when I had things arranged differently and never cleaned them up. Thanks for the detailed reading

python/lsst/meas/astrom/fitAffineWcs.py
natelust: Yeah I thought about this too. Most of the time seems to be "within this function" so I suspect it is things like this. However it was so much faster I didn't know if it was worth opening up another repo and adding an interface which is pretty much just for this. I would be happy to do so if we thought it worth it.

TallJimbo: Let's just get this in.  Doing more could easily take a lot longer than this ticket has so far.

PR #185, opened 2023-08-26
DM-39796: Update task config defaults to LSST values

parejkoj:


python/lsst/meas/astrom/astrometry.py
```diff
@@ -75,7 +75,7 @@ class AstrometryConfig(RefMatchConfig):
         dtype=bool,
         doc=("If True then a rough zeropoint will be computed from matched sources "
              "and outliers will be rejected in the iterations."),
-        default=False,
+        default=True,
```
cmsaunders, line 78: I see that this now matches the LSST value and that the corresponding config setting has been removed in `obs_lsst`. However, do you want to add `doMagnitudeOutlierRejection=False` to `obs_subaru`, or are you purposely changing the behavior there?

parejkoj: I have overridden it in obs_subaru, as part of a commit that tries to ensure HSC processing is unchanged after this ticket. It also has a note that they should review those new overrides, as they are likely unnecessary.

include/lsst/meas/astrom/matchOptimisticB.h
```diff
@@ -88,7 +88,7 @@ struct MatchOptimisticBControl {
     MatchOptimisticBControl()
             : refFluxField("r_flux"),
               sourceFluxField("slot_ApFlux_instFlux"),
-              numBrightStars(100),
+              numBrightStars(150),
```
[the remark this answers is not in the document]
parejkoj: Yes, it's unclear why LSST used a different value, and I don't think there's any obvious a priori reason to take the LSST value over the HSC one.

PR #187, opened 2023-10-20
DM-40942: Switch AstrometryTask to use ScienceSourceSelector

parejkoj:


python/lsst/meas/astrom/astrometry.py
```diff
@@ -87,14 +86,21 @@ class AstrometryConfig(RefMatchConfig):
...
-
-        # Note that if the matcher is MatchOptimisticBTask, then the default
-        # should be self.sourceSelector['matcher'].excludePixelFlags = False
-        # However, there is no way to do this automatically.
+        self.sourceFluxType = "Psf"
+        # Configured to match the deprecated "matcher" selector: isolated,
+        # SN > 40, some bad flags, valid centroids.
+        self.sourceSelector["science"].doSignalToNoise = True
+        self.sourceSelector["science"].signalToNoise.minimum = 40
+        self.sourceSelector['science'].signalToNoise.fluxField = f"slot_{self.sourceFluxType}Flux_instFlux"
```
leeskelvin, line 94: In keeping with the rest of the script, these should probably be `"`.

PR #188, opened 2024-03-14
DM-43332: Improve match_probabilistic performance

taranu:


python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -522,27 +517,39 @@ def match(
...
+        if coords_spherical:
+            match_dist_max = config.match_dist_max/3600.
+            with Matcher(coords_target[0], coords_target[1]) as matcher:
+                idxs_target_select = matcher.query_knn(
+                    coords_ref[0], coords_ref[1],
+                    distance_upper_bound=match_dist_max,
+                    k=config.match_n_max,
+                )
+        # Call scipy for non-spherical case
+        # The spherical case won't trigger, but the implementation is left for comparison, if needed
```
enourbakhsh, line 538: Do you mean "The **non-spherical** case won't trigger"?

enourbakhsh: How do you anticipate such comparisons becoming necessary?

python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -601,24 +632,28 @@ def match(
...
-                    except Exception as error:
-                        # Can't foresee any exceptions, but they shouldn't prevent
-                        # matching subsequent sources
-                        exceptions[index_row] = error
+                if not any(chisq_good):
+                    continue
+                try:
+                    chisq_sum = np.zeros(n_found, dtype=float)
+                    chisq_sum[chisq_good] = np.nansum(chi[chisq_good, :] ** 2, axis=1)
+                    idx_chisq_min = np.nanargmin(chisq_sum / n_finite)
```
enourbakhsh, line 640: Are we always safeguarded against division by zero here?

python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -388,10 +381,11 @@ def columns_in_target(self) -> Set[str]:
         default=10,
         optional=True,
         doc='Maximum number of spatial matches to consider (in ascending distance order).',
+        check=lambda x: x >= 1,
     )
     match_n_finite_min = pexConfig.Field(
         dtype=int,
-        default=3,
+        default=2,
```
enourbakhsh, line 388: Maybe add a `check` here to specify the valid domain?

enourbakhsh (commented):
It's probably not on the agenda for this ticket, but it would make sense to change single quotes to double quotes for strings later. Below are my other comments.

python/lsst/meas/astrom/matcher_probabilistic.py
taranu: No, I mean that the spherical case won't use scipy but will use smatch instead. So the `_radec_to_xyz if coords_spherical else np.vstack` on line 542 will never actually call `_radec_to_xyz`.

I anticipate most users will match on spherical coordinates and so the else block won't get called much.

python/lsst/meas/astrom/matcher_probabilistic.py
taranu: As long as n_finite_min > 0, yes.

enourbakhsh: Yes, that's why I suggested you add `check` in `match_n_finite_min = pexConfig.Field(...)`.

PR #192, opened 2024-06-13
DM-44805: Fix setting of astromOffsetStd metric in meas_astrom

laurenam:


tests/test_astrometryTask.py
```diff
@@ -92,6 +92,10 @@ def doTest(self, pixelsToTanPixels):
         # Test that the sources used in the fit are flagged in the catalog.
         self.assertEqual(sum(sourceCat["calib_astrometry_used"]), len(results.matches))
 
+        self.assertNotEqual(
+            self.exposure.metadata["SFM_ASTROM_OFFSET_MEAN"], self.exposure.metadata["SFM_ASTROM_OFFSET_STD"]
```
parejkoj, line 96: I'm also not keen on this kind of test (at the very least, please add a comment for why you're doing this). I wonder if you can just directly calculate the values for this from the input data and test against those?

laurenam: I don't know how to easily compute the values based on inputs, but I added a detailed comment about the check and its potential pitfalls (so, on the off chance it does somehow end up causing problems, there will be some context to go on if it needs adjusting/eliminating).

PR #196, opened 2024-08-19
DM-45782: Make selected refcat contiguous if necessary

parejkoj:


erinleighh (approved):
Fix looks good and works great, thank you!

PR #197, opened 2024-09-16
DM-44159: Replace pandas.DataFrame usage with astropy.Table in Match(er)Probabilistic

taranu:


python/lsst/meas/astrom/match_probabilistic_task.py
```diff
@@ -71,17 +72,21 @@ def _apply_select_bool(
...
         if select_additional:
             if selection is None:
                 selection = np.ones(len(catalog), dtype=bool)
             for column in columns_true:
                 # This is intended for boolean columns, so the behaviour for non-boolean is not obvious
                 # More config options and/or using a ConfigurableActionField might be best
-                values = catalog[column].values
+                values = catalog[column]
+                if is_pd:
+                    values = values.values
```
erykoff, line 85: Pandas really forces into ugly constructions.  I don't love the look of this, though.  `values.values` is confusing.
Maybe something like
`catalog_values = catalog[column] if not is_pd else catalog[column].values`

python/lsst/meas/astrom/match_probabilistic_task.py
```diff
@@ -71,17 +72,21 @@ def _apply_select_bool(
...
                 selection &= (np.isfinite(values) & (values != 0))
             for column in columns_false:
-                selection &= (catalog[column].values == 0)
+                values = catalog[column]
+                selection &= ((values.values if is_pd else values) == 0)
```
erykoff, line 89: This could be `selection &= ((catalog[column] if not is_pd else catalog[column].values) == 0)` though that is a mouthful.

python/lsst/meas/astrom/match_probabilistic_task.py
```diff
@@ -124,20 +129,28 @@ def match(
...
+        catalog_out_target : `pandas.DataFrame` | `astropy.table.Table`
             Reference matched catalog with indices of target matches.
         """
         if logger is None:
             logger = self.log
 
+        is_ref_pd = isinstance(catalog_ref, pd.DataFrame)
+        is_target_pd = isinstance(catalog_target, pd.DataFrame)
+        if is_ref_pd or is_target_pd:
+            logger.warning("pandas DataFrame inputs are deprecated")
```
erykoff, line 143: Is there a deprecation ticket?  There needs to be a ticket, and the warning should use a python deprecation warning, probably following https://developer.lsst.io/stack/deprecating-interfaces.html#package-deprecation if we can't use the sphinx stuff on a call.

python/lsst/meas/astrom/match_probabilistic_task.py
```diff
@@ -124,20 +129,28 @@ def match(
...
 
         if config.column_ref_order is None:
-            flux_tot = np.nansum(
-                catalog_ref.loc[:, config.columns_ref_flux].values, axis=1
+            fluxes = (
+                catalog_ref.loc[:, config.columns_ref_flux].values
+                if is_ref_pd else
+                [catalog_ref[key] for key in config.columns_ref_flux]
             )
+            flux_tot = np.nansum(fluxes, axis=1 if is_ref_pd else 0)
```
erykoff, line 153: Is this needed because the pandas selection leaves an extra dimension of length `1`?  If this is the case would `np.squeeze()` help in the pandas case above?

python/lsst/meas/astrom/match_probabilistic_task.py
```diff
@@ -215,8 +228,14 @@ def run(
...
-        catalog_ref.reset_index(inplace=True)
-        catalog_target.reset_index(inplace=True)
+        is_ref_pd = isinstance(catalog_ref, pd.DataFrame)
+        is_target_pd = isinstance(catalog_target, pd.DataFrame)
+        if is_ref_pd:
+            catalog_ref.reset_index(inplace=True)
+        if is_target_pd:
+            catalog_target.reset_index(inplace=True)
+        if is_ref_pd or is_target_pd:
+            self.log.warning("pandas DataFrame inputs are deprecated")
```
erykoff, line 238: As above, this should be a proper deprecation warning.

python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -208,6 +209,16 @@ def format_catalogs(
...
         """
+        logger = kwargs.get("logger", logger_default)
+        is_ref_pd = isinstance(catalog_ref, pd.DataFrame)
+        is_target_pd = isinstance(catalog_target, pd.DataFrame)
+        if is_ref_pd:
+            catalog_ref = astropy.table.Table.from_pandas(catalog_ref)
+        if is_target_pd:
+            catalog_target = astropy.table.Table.from_pandas(catalog_target)
+        if is_ref_pd or is_target_pd:
+            logger.warning("pandas DataFrame inputs are deprecated")
```
erykoff, line 220: And here for a proper deprecation warning.

python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -486,6 +497,15 @@ def match(
...
             logger = logger_default
 
+        is_ref_pd = isinstance(catalog_ref, pd.DataFrame)
+        is_target_pd = isinstance(catalog_target, pd.DataFrame)
+        if is_ref_pd:
+            catalog_ref = astropy.table.Table.from_pandas(catalog_ref)
+        if is_target_pd:
+            catalog_target = astropy.table.Table.from_pandas(catalog_target)
+        if is_ref_pd or is_target_pd:
+            logger.warning("pandas DataFrame inputs are deprecated")
```
erykoff, line 507: And here for a deprecation warning.

python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -509,9 +529,9 @@ def match(
         # Also, this is done on the original dataframe as it's harder to accomplish
         # just with a recarray
         column_order = (
-            catalog_ref.loc[ref.extras.select, config.column_ref_order]
+            catalog_ref[config.column_ref_order][ref.extras.select]
             if config.column_ref_order is not None else
-            np.nansum(catalog_ref.loc[ref.extras.select, config.columns_ref_flux], axis=1)
+            np.nansum([catalog_ref[col][ref.extras.select] for col in config.columns_ref_flux], axis=0)
```
erykoff, line 534: Comment above refers to a dataframe...

tests/test_match_probabilistic_task.py
```diff
@@ -26,8 +26,8 @@
 import lsst.afw.geom as afwGeom
 from lsst.meas.astrom import ConvertCatalogCoordinatesConfig, MatchProbabilisticConfig, MatchProbabilisticTask
 
+import astropy.table
```
erykoff, line 29: Unfortunately, we need to test the deprecated pandas usage as well, but catching the deprecation warnings.

erykoff (approved):
Probably not worth fixing, but I find it a bug unusual to have dataframe listed before astropy table in the type defs when the astropy table is the primary one.  But this isn't really user visible, so it's just a style thing.

erykoff:
Also I would retitle the PR to make it clear which tasks are getting changed.

PR #199, opened 2024-10-02
DM-46567: Fix bugs in match disambiguation chi calculation

taranu:
The updated unit test should catch future regressions.

This was evidently not tested well enough in DM-44159.


jtmccann (approved):
Looks good to me.

PR #209, opened 2025-06-09
DM-46523: Remove pandas support in probabilistic matcher

taranu:


python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -79,8 +78,8 @@ class CatalogExtras:
 
     Parameters
     ----------
-    catalog : `pandas.DataFrame`
-        A pandas catalog to store extra information for.
+    catalog : `astropy.table.Table`
+        A Table to store extra information for.
```
nevencaplar, line 82: I know this is not the main point of the ticket, but the description of the catalog here is very unclear. A Table to store extra information for? It is ok not to change.

PR #219, opened 2026-01-15
DM-53825: LSSTCam DRP Error in calibrateImage: cannot access local variable 'refColorDelta' where it is not associated with a value

laurenam:


isullivan (approved):
One minor suggestion.

PR #221, opened 2026-02-16
DM-54159: Update use of butler-stored detector pointing and regions in single frame processing

laurenam:


python/lsst/meas/astrom/astrometry.py
```diff
@@ -292,12 +313,20 @@ def solve(self, exposure, sourceCat):
...
-            filterName=band,
-            epoch=epoch,
-        )
+        if exposure_region is not None:
+            loadResult = self.refObjLoader.loadRegion(
+                region=exposure_region,
+                filterName=band,
+                epoch=epoch,
+                wcsForCentroids=exposure.wcs,
+            )
```
TallJimbo, line 322: Should we warn or raise here if the `refObjLoader` config is set to add padding, since that's now being ignored?

TallJimbo (approved):

python/lsst/meas/astrom/astrometry.py
laurenam: I've gone with a `warning` along the lines of:
```
if self.refObjLoader.config.pixelMargin > 0:
    self.log.warning("Note that the astrometry_ref_loader.pixelMargin (currently "
                     "set to %d) is ignored when loading the reference catalog "
                     "with an exposure_region (i.e. the region is used as is, with "
                     "no additional padding).", self.refObjLoader.config.pixelMargin)
```

So, e.g., the following would get printed:
`lsst.calibrateImage.astrometry WARNING: Note that the astrometry_ref_loader.pixelMargin (currently set to 800) is ignored when loading the reference catalog with an exposure_region (i.e. the region is used as is, with no additional padding).`

I think a `raise` would be a bit much given how many set-but-unused configs we have in the wild...feel free to shout if you disagree as I'll run another "just-in-case" Jenkins before merging.

TallJimbo: That sounds reasonable.  If we see the warning too much we can adjust course.

PR #222, opened 2026-04-03
DM-51163: Fix bugs in/improve probabilistic matcher

taranu:


python/lsst/meas/astrom/matcher_probabilistic.py
```diff
@@ -351,12 +351,12 @@ def columns_in_target(self) -> Set[str]:
...
     columns_target_select_true = pexConfig.ListField(
         dtype=str,
-        default=('detect_isPrimary',),
+        default=[],
         doc='Target table columns to require to be True for selecting sources',
     )
     columns_target_select_false = pexConfig.ListField(
         dtype=str,
-        default=('merge_peak_sky',),
+        default=[],
```
sr525, line 367: Why are you removing this default?

taranu: The matching will be done to `object` by default, so these columns won't even exist. If someone wants to match to `object_all` they can put the overrides back.

PR #223, opened 2026-05-10
DM-54866: Add a rough estimate of the per-detector reference catalog source density

laurenam:


erykoff (dismissed):
Sigh; the mix of 🐍 and 🐫 here.  But it was there before and now even more!

erykoff (approved):

